PR #121, opened 2023-02-08
Jaelynlitz/get aer num

jaelynlitz:
Add port of get_aer_num from [ndrop.F90](https://github.com/eagles-project/e3sm_mam4_refactor/blob/refactor-maint-2.0/components/eam/src/physics/cam/ndrop.F90#L1669) 

TODO:
- [X] get dummy test to not seg fault
- [X] write a more robust test

### DIFF
--- a/src/mam4xx/ndrop.hpp
+++ b/src/mam4xx/ndrop.hpp
@@ -1,56 +1,52 @@
-subroutine get_aer_num(imode, istart, istop, state_q, cs, vaerosol, qcldbrn1d_num, &!in
-           naerosol) !out
+#ifndef MAM4XX_NDROP_HPP
+#define MAM4XX_NDROP_HPP
 
-  use modal_aero_data, only:numptr_amode
+#include <haero/aero_species.hpp>
+#include <haero/constants.hpp>
+#include <haero/atmosphere.hpp>
+#include <haero/math.hpp>
 
-  ! input arguments
-  integer,  intent(in) :: imode        ! mode index
-  integer,  intent(in) :: istart       ! start column index (1 <= istart <= istop <= pcols)
-  integer,  intent(in) :: istop        ! stop column index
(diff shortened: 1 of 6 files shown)


jeff-cohere:
Hey @jaelynlitz . Thanks for adding me. This looks pretty good in general. I'll take a closer look at it tomorrow morning to make sure I understand what's supposed to happen.

src/mam4xx/ndrop.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <mam4xx/mam4_types.hpp>
+#include <mam4xx/utils.hpp>
+
+using Real = haero::Real;
+
+namespace mam4 {
+
+KOKKOS_INLINE_FUNCTION
+void get_aer_num(const Diagnostics &diags, const Prognostics &progs,
+                 const Atmosphere &atm, int mode_idx, int icol,
```
jeff-cohere, line 20: It seems to me that `icol` is being used for the vertical level index (e.g. `atm.temperature(icol)` is the temperature of the atmospheric column of air at the vertical level `icol`). Our convention for this variable is `k`, whereas `icol` (or `i`) usually refers to which column we're looking at. Can we rename this variable to `k`?

pbosler (approved):

jaelynlitz:
As per discussion with @jeff-cohere - the GPU aspect of the unit test for this function is not yet functional and hard to debug/finalize without getting a better idea of the context where this function will be called. I have left TODO comments in areas that will need to be revisited. I have disabled this test on GPU for now, but the tests currently pass on CPU.

jeff-cohere:
Your exclusion of the ndrop test will work for the PNNL CI system, but we'll still see failures if we build in other GPU environments. Can you instead comment out the unit test in CMakeLists.txt so we don't get distracted by this? We'll return to fix this when we need it for a ported parameterization.

Thanks!

jeff-cohere (approved):
Thanks, @jaelynlitz . This is ready to merge.

PR #133, opened 2023-02-27
Move max_coagpair

pressel:
This PR ties up some loose ends from coagulation and aging. In particular, the variable max_coagpair which defines the maximum number of coagulation pairs is now defined in aero_config.hpp. Additionally,  a unit test is added for both the aging and coagulation classes to guarantee that max_coagpair is not changed to be inconsistent with the value expected in both of those classes.


jeff-cohere (approved):
:+1: Thanks Kyle!

PR #134, opened 2023-03-02
Remove redundant rules (bc of push mirror)

jaelynlitz:
These rules are no longer checked because of how the push mirror triggers the gitlab pipelines


jeff-cohere (approved):
It seems like the CI system is still running without this code, so looks good!

PR #136, opened 2023-03-08
Adds a `make memcheck` target that runs tests with Valgrind.

jeff-cohere:
I considered having our CI tests run with Valgrind/memcheck, but some of the cross-validation tests take a very long time with this. If running tests with Valgrind is interesting to us, we might consider altering some of the larger cross-validation tests, like `run_vehkamaki2002_fig8`, which generates a contour plot of the nucleation rate as a function of 2 variables.

### DIFF
--- a/cmake/add_memcheck_target.cmake
+++ b/cmake/add_memcheck_target.cmake
@@ -1,0 +1,25 @@
+# This macro adds a "make memcheck" target that runs Valgrind on all tests.
+# It only works on Linux.
+macro(add_memcheck_target)
+  if (UNIX AND NOT APPLE)
+    find_program(VALGRIND valgrind)
+    if (NOT VALGRIND MATCHES "-NOTFOUND")
+      message(STATUS "Valgrind found. Enabling `make memcheck`")
+      set(VALGRIND_FOUND 1)
+      set(CTEST_MEMORYCHECK_COMMAND ${VALGRIND})
+      # Add "--gen-suppressions=all" to MEMORYCHECK_COMMAND_OPTIONS to generate
+      # suppressions for Valgrind's false positives. The suppressions show up
+      # right in the MemoryChecker.*.log files.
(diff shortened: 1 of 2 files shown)


jeff-cohere:
Sigh. Looks like I've messed up EKAT's hyper-brittle installation process in my latest Haero PR.

odiazib (approved):

cmake/add_memcheck_target.cmake
```diff
@@ -0,0 +1,25 @@
...
+
+      # make memcheck target
+      add_custom_target(memcheck ctest -T memcheck -j USES_TERMINAL)
+    else()
+      set(VALGRIND_FOUND 0)
+    endif()
+  else()
+    # Valgrind doesn't work on Macs.
+    set(VALGRIND_FOUND 0)
+  endif()
```
CameronRutherford, line 23: This adds a custom target, but does it run any tests? Are these memcheck tests meant to be running somewhere in GitHub actions or PNNL actions?

I couldn't find anything in the logs from looking at it :(

jeff-cohere: This adds a custom target and teaches `ctest` how to run with the `-T memcheck` flag. Because Valgrind slows everything way down, we don't actually use these tests in CI, and I wouldn't suggest we do unless we do some pruning. So for now, this is a "run on your machine" target.

CameronRutherford: We couldn't have say one GitHub action exclusively run Valgrind tests for one build type?

My experience with checks that we hope people remember to run (as well as run on a non-MacOS) is that they often forget.

jeff-cohere: We could do that, but not with our current batch of tests--they just take too long. I don't think it would be difficult to prune the validation tests so they'd run faster, and once we do that, we can add a memcheck test.

Issue #137, opened 2023-03-13
Bump Required CMake to v3.17?


mjs271:
I wanted to get the team's thoughts on moving to a minimum required CMake 3.17.

Currently, we require CMake 3.12, but 3.17 has some nice functionality. My immediate interest is in `ZIP_LISTS` that provides the same functionality as a python-style zipped list. I am interested in using it to parallel iterate over validation tests when a single default passing tolerance does not work for all tests. This uses the `add_test()` [syntax](https://github.com/eagles-project/mam_x_validation/blob/2dda50c75297cecb7ad2a051087890ab3ce69f96/scripts/compare_mam4xx_mam4.py#L40) introduced by Oscar. E.g., 
```
foreach(input tol IN ZIP_LISTS TEST_LIST ERROR_THRESHOLDS)
[...]
add_test(validate_${input} python3 compare_mam4xx_mam4.py mam4xx_${input}.py mam_${input}.py True ${tol})
```

- **Weaver**: 3.12 is module default but available versions range from 3.6 to 3.24.
- **Blake**: 3.22 is module default but available versions range from 3.4 to 3.25.
- **Chrysalis**: 3.20 pre-loaded by default, but modules for 3.19 and 3.24.
- **Compy**: has 3.19 (*thanks, Jeff!*)
- [ ] **Mappy**: *I don't have an acccount--would someone mind checking?*

@jeff-cohere @pressel @overfelt @pbosler @singhbalwinder @odiazib

pbosler:
I have no objection, but we should make sure whatever we choose is compatible with EKAT.   EKAT drives the c++ standard, for example.

singhbalwinder:
I am also fine with the new version.

jeff-cohere:
EKAT currently only requires CMake 3.3 (as does EAMxx), but I'm skeptical about whether either of those codes would actually build with such an ancient version of CMake at this point.

PR #138, opened 2023-03-15
move `cmake_minimum_required` to v3.17

mjs271:
In reference to issue #137, this PR updates the cmake minimum required version to v3.17 and also updates the documentation in the `README` and `docs/installation.md`.


jeff-cohere (approved):

pbosler (approved):
It looks like the auto test workflow automatically satisfies the new requirement (the default cmake is already >= 3.17), which is nice.

PR #143, opened 2023-03-21
Fixing a possible race condition in uploading code coverage reports.

jeff-cohere:
This PR makes it so only one of our builds attempts to upload code coverage reports.


mjs271:
> Well, shoot. It looks like codecov is just being flaky! Regardless, it's probably a good idea for us to only upload one code coverage report.

All forward progress counts! 🙂

mjs271 (approved):

PR #145, opened 2023-03-24
Jaelynlitz/explmix

jaelynlitz:
port of function explmix from https://github.com/eagles-project/e3sm_mam4_refactor/blob/refactor-maint-2.0/components/eam/src/physics/cam/ndrop.F90#L1216

### DIFF
--- a/.github/workflows/auto_test.yml
+++ b/.github/workflows/auto_test.yml
@@ -88,7 +88,7 @@ jobs:
     - name: Building MAM4xx (${{ matrix.build-type }}, ${{ matrix.fp-precision }} precision)
       run: |
         cd build
-        make -j
+        make
 
     - name: Installing Python 3
       uses: actions/setup-python@v4
--- a/src/mam4xx/ndrop.hpp
+++ b/src/mam4xx/ndrop.hpp
@@ -73,34 +73,26 @@ void explmix(
 
   Kokkos::parallel_for(
       "compute q per level", nlev, KOKKOS_LAMBDA(const int k) {
-        //  the parallel for is causing the last q to be .1 shy of the correct
-        //  answer
-        //  ?????????????
         int kp1 = min(k + 1, nlev - 1);
         int km1 = max(k - 1, top_lev);
 
         // the qactold*(1-overlap) terms are resuspension of activated material
 
         if (is_unact) {
-          printf("q = %f\n", q(k));
           q(k) = qold(k) +
                  (dt * (-src(k) +
                         (ekkp(k) * (qold(kp1) - qold(k) +
                                     (qactold(kp1) * (1 - overlapp(k))))) +
                         (ekkm(k) * (qold(km1) - qold(k) +
                                     (qactold(km1) * (1 - overlapm(k)))))));
-          printf("q = %f\n", q(k));
         } else {
-          printf("q = %f\n", q(k));
           q(k) = qold(k) +
                  (dt *
                   (src(k) + (ekkp(k) * ((overlapp(k) * qold(kp1)) - qold(k))) +
                    (ekkm(k) * ((overlapm(k) * qold(k)) - qold(k)))));
-          printf("q = %f\n", q(k));
         }
         // force to non-negative
         q(k) = max(q(k), 0);
-        printf("q = %f\n", q(k));
       });
 }
 
--- a/src/mam4xx/ndrop.hpp
+++ b/src/mam4xx/ndrop.hpp
@@ -19,33 +19,84 @@ namespace mam4 {
 //  come back when this function is being used in a ported parameterization
 KOKKOS_INLINE_FUNCTION
 void get_aer_num(const Diagnostics &diags, const Prognostics &progs,
-                 const Atmosphere &atm, const int mode_idx, const int k,
-                 Real naerosol[AeroConfig::num_modes()]) {
+                const Atmosphere &atm, const int mode_idx, const int k,
+                Real naerosol[AeroConfig::num_modes()]) {
 
-  Real rho =
-      conversions::density_of_ideal_gas(atm.temperature(k), atm.pressure(k));
-  Real vaerosol = conversions::mean_particle_volume_from_diameter(
-      diags.dry_geometric_mean_diameter_total[mode_idx](k),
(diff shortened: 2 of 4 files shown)


src/mam4xx/ndrop.hpp
```diff
@@ -47,5 +47,51 @@ void get_aer_num(const Diagnostics &diags, const Prognostics &progs,
...
+    Real qold[4], // number / mass mixing ratio from previous time step [# or kg
+                  // / kg]
+    Real dt,      // time step [s]
+    bool is_unact,  // true if this is an unactivated species
+    Real qactold[4] // optional: number / mass mixing ratio of ACTIVATED species
+                    // from previous step *** this should only be present if the
+                    // current species is unactivated number/sfc/mass
+) {
+
+  int top_lev = 0; //??
```
jaelynlitz, line 71: I assume we start indexing at 0?

src/mam4xx/ndrop.hpp
```diff
@@ -47,5 +47,51 @@ void get_aer_num(const Diagnostics &diags, const Prognostics &progs,
   naerosol[mode_idx] = max(naerosol[mode_idx], vaerosol * num2vol_ratio_max);
   naerosol[mode_idx] = min(naerosol[mode_idx], vaerosol * num2vol_ratio_min);
 }
+
+KOKKOS_INLINE_FUNCTION
+void explmix(
+    int pver,     // number of levels
+    Real q[4],    // number / mass mixing ratio to be updated [# or kg / kg]
```
jaelynlitz, line 54: I know that these should not be the number 4, but placeholder to check the tests while we discuss the solution.

jaelynlitz (commented):
Recommended to put together a validation test driver as well

src/mam4xx/ndrop.hpp
jeff-cohere: Yes, that's correct. And the indexing goes from the top of the atmosphere to the surface of the earth, for reasons related to the "dycore" (dynamical core), which moves things around between columns.

src/mam4xx/ndrop.hpp
```diff
@@ -47,5 +47,51 @@ void get_aer_num(const Diagnostics &diags, const Prognostics &progs,
   naerosol[mode_idx] = max(naerosol[mode_idx], vaerosol * num2vol_ratio_max);
   naerosol[mode_idx] = min(naerosol[mode_idx], vaerosol * num2vol_ratio_min);
 }
+
+KOKKOS_INLINE_FUNCTION
+void explmix(
+    int pver,     // number of levels
```
jeff-cohere, line 53: Let's call this `nlev`. `pver` is a historical name, and I still have no idea what the `p` stands for.

src/mam4xx/ndrop.hpp
jeff-cohere: How are these arrays sized in the Fortran code?

jeff-cohere (changes requested):
Looks good! Left a couple of suggestions and a question about the 4-sized arrays. Also, you're right that we should set up an `ndrop` validation driver that calls this function in addition to others, if we haven't got one of those yet. But that can be done in a subsequent PR.

pbosler:
>we should set up an ndrop validation driver that calls this function in addition to others

This function can actually be verified -- I believe it should solve the 1D heat equation with 2nd order accuracy.   We can build a verification test around that (in a separate PR, perhaps -- if so, we should make an issue to remember to do it).

jeff-cohere:
@pbosler , is this true, given the conversation you had with Dick Easter about this diffusion operator not corresponding exactly to a finite difference representation? Or am I misremembering?

pbosler:
You're remembering right.  Still, in principle, it's a verification test that we can implement.   Maybe that's a good reason to save that test for a separate PR.

src/mam4xx/ndrop.hpp
jaelynlitz: They are sized based on `pver`: 
```
real(r8), intent(out) :: q(pver)
```

jaelynlitz:
Thanks for the discussion and review @jeff-cohere and @pbosler, I'll open an issue for writing a validation/verification for ndrop and explmix

src/mam4xx/ndrop.hpp
jeff-cohere: Okay, that makes sense. But unfortunately, I don't think this will work as is, since `q` is an array of size 4, and our typical `nlev` is 72. I wonder if we need to use `ColumnView`s instead of arrays for this function.

jeff-cohere: We might have to take a look at the function/subroutine that calls this one to figure out how best to pass the data in.

jeff-cohere:
Okay, I pushed a commit with some minor fixes (plus a reformat, plus a workaround for a compiler bug in GitHub's CI environment). The only remaining thing is to add a `ThreadTeam` argument to `explmix` and to have the test call it using a team dispatch like the one done [here](https://github.com/eagles-project/mam4xx/blob/main/src/tests/mam4_aging_unit_tests.cpp#L81). This allows us to test that the thread team passed to `explmix` by the caller exhibits the right behavior.

jeff-cohere (approved):
Looks great! Nice work. If you think it's ready, go ahead and remove the Draft status.

jaelynlitz:
Ready for final review @pbosler @pressel

jeff-cohere:
Hey all. Last call for reviews! If no one has any objections to what's in this PR, let's merge it.

CameronRutherford (approved):
1 minor question

.github/workflows/auto_test.yml
```diff
@@ -88,7 +88,7 @@ jobs:
     - name: Building MAM4xx (${{ matrix.build-type }}, ${{ matrix.fp-precision }} precision)
       run: |
         cd build
-        make -j
+        make
```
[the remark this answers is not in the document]
CameronRutherford: Very odd... Not sure it's worth anyone's time to debug this anyway since we can't reproduce locally, but perhaps some CMake wizardry would fix things. It's also possible constraining the number of parallel jobs within the actions runner to some nominal n<64 or something would fix it, but again, not worth the time testing.

PR #146, opened 2023-03-25
ndrop maxsat

jaelynlitz:
port of function maxsat from ndrop: https://github.com/eagles-project/e3sm_mam4_refactor/blob/refactor-maint-2.0/components/eam/src/physics/cam/ndrop.F90#L1417

### DIFF
--- a/src/mam4xx/ndrop.hpp
+++ b/src/mam4xx/ndrop.hpp
@@ -18,30 +18,21 @@ namespace mam4 {
 class NDrop {
 
 public:
-
-    //TODO: put this inside ndrop_init
-    // abdul-razzak functions of width
-    //static Real bizarro1[AeroConfig::num_modes()];
-    // abdul-razzak functions of width
-    //static Real bizarro2[AeroConfig::num_modes()];
-
 };
-
 
 namespace ndrop {
 
-//static ColumnView bizarro1;
-//static ColumnView bizarro2;
+// abdul-razzak functions of width
+static Real f1[AeroConfig::num_modes()];
+static Real f2[AeroConfig::num_modes()];
 
-
-KOKKOS_INLINE_FUNCTION 
+KOKKOS_INLINE_FUNCTION
 void ndrop_init() {
-    //bizarro1 = ColumnView("bizarro1", AeroConfig::num_modes());
-    //bizarro2 = ColumnView("bizarro2", AeroConfig::num_modes());
-    for(int m = 0; m < AeroConfig::num_modes(); m++) {
-        //bizarro1(m) = 0.5 * haero::exp(2.5 * haero::log(modes(m).mean_std_dev) * haero::log(modes(m).mean_std_dev));
-       // bizarro2(m) = 1.0 + 0.25 * haero::log(modes(m).mean_std_dev);
-    }
+  for (int m = 0; m < AeroConfig::num_modes(); m++) {
+    f1[m] = 0.5 * haero::exp(2.5 * haero::log(modes(m).mean_std_dev) *
+                             haero::log(modes(m).mean_std_dev));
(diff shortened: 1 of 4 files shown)


src/mam4xx/ndrop.hpp
```diff
@@ -15,6 +15,21 @@ using Real = haero::Real;
 
 namespace mam4 {
 
+namespace ndrop {
+//TODO: put this inside ndrop_init
+// abdul-razzak functions of width
+Real bizarro1[AeroConfig::num_modes()];
+// abdul-razzak functions of width
+Real bizarro2[AeroConfig::num_modes()];
```
jaelynlitz, line 23: these are f1 and f2, but getting a weird error that mam4::f1 already exists? 
```
libmam4xx_tests.a(kohler_verification.cpp.o):(.bss+0x0): multiple definition of `mam4::ndrop::bizarro1'
CMakeFiles/kohler_unit_tests.dir/kohler_unit_tests.cpp.o:(.bss+0x0): first defined here
libmam4xx_tests.a(kohler_verification.cpp.o): In function `__gnu_cxx::char_traits<char>::eq(char const&, char const&)':
/people/litz372/mam4/mam4xx/src/tests/kohler_verification.cpp:17: multiple definition of `mam4::ndrop::bizarro2'
CMakeFiles/kohler_unit_tests.dir/kohler_unit_tests.cpp.o:/people/litz372/mam4/mam4xx/haero_cpu_double_debug/include/spdlog/pattern_formatter-inl.h:160: first defined here
collect2: error: ld returned 1 exit status
```

src/mam4xx/ndrop.hpp
```diff
@@ -15,6 +15,21 @@ using Real = haero::Real;
...
+
+KOKKOS_INLINE_FUNCTION 
+void ndrop_init() {
+    for(int m = 0; m < AeroConfig::num_modes(); m++) {
+        bizarro1[m] = 0.5 * haero::exp(2.5 * haero::log(modes(m).mean_std_dev) * haero::log(modes(m).mean_std_dev));
```
jaelynlitz, line 28: Am I right in deducing that `sigmag_amode(m)` is `modes(m).mean_std_dev`?

src/mam4xx/ndrop.hpp
```diff
@@ -93,5 +108,48 @@ void explmix(
...
+            ) {
+
+    Real sum = 0;
+    Real g1, g2;
+    bool weak_forcing = true; // whether forcing is sufficiently weak or not
+
+    for(int m = 0; m < nmode; m++) {
+        if(zeta > 1e5 * eta[m] || smc[m] * smc[m] > 1e5 * eta[m]) {
+            // weak forcing. essentially none activated
+            smax = 1e-20;
```
jaelynlitz, line 130: not 100% positive that this e syntax works in c++

jeff-cohere: Do you mean the lack of parenthesis in the `if` expression? I think logical operators have lower precedence than ordering operators. Of course, adding parens usually improves clarity.

src/mam4xx/ndrop.hpp
jeff-cohere: If you want to have these be global variables, you should declare them as `static` so that different `.cpp` files don't all define them separately. That's what's happening to produce this error.

It's probably better not to have these live in the header file, or to attach them to an instance like your comment suggests.

jeff-cohere (commented):
Took a quick look and left some comments.

src/mam4xx/ndrop.hpp
jaelynlitz: Ah, perhaps something to live in the NDrop class?

pbosler (approved):

PR #155, opened 2023-04-05
Add conversion factor of 1e-3 to `conversions.hpp` to account for change in units

### LINKED ISSUE #152: Issue with units in vmr $\leftrightarrow$ mmr conversion calculations
It appears we are off by a factor of 1000 in our volume mixing ratio $\leftrightarrow$ mass mixing ratio conversion.

In [`mam4xx/conversions.hpp`](https://github.com/eagles-project/mam4xx/blob/2101c5fa5392fd4f77fe08cab6ab24e5f9df446d/src/mam4xx/conversions.hpp#L51-L72) the units are all **kilo-** and appear to line up properly:
```
/// Given a molar mixing ratio (vmr) for a species or mixture
/// [kmol species/kmol dry air], computes and returns a mass mixing ratio
/// [kg species/kg dry air] based on its molecular weight.
/// @param [in] vmr The molar mixing ratio of the species/mixture [kmol/kmol
/// air]
/// @param [in] molecular_wt The molecular weight of the species/mixture
/// [kg/kmol]
KOKKOS_INLINE_FUNCTION Real mmr_from_vmr(Real vmr, Real molecular_wt) {
  const auto mw_dry_air = Constants::molec_weight_dry_air;
  return vmr * molecular_wt / mw_dry_air;
}

/// Given a mass mixing ratio (mmr) for a species or mixture [kg species/kg
/// dry air], computes and returns a molar mixing ratio [kmol species/k dry air]
/// based on its molecular weight.
/// @param [in] mmr The mass mixing ratio of the species/mixture [kg/kg dry air]
/// @param [in] molecular_wt The molecular weight of the species/mixture
/// [kg/kmol]
KOKKOS_INLINE_FUNCTION Real vmr_from_mmr(Real mmr, Real molecular_wt) {
  const auto mw_dry_air = Constants::molec_weight_dry_air;
  return mmr * mw_dry_air / molecular_wt;
}
```
However the value of `Constants::molec_weight_dry_air` from [`haero/constants.hpp`](https://github.com/eagles-project/haero/blob/825bde24eb3c6b204e44cd7ee9104451888e6a31/haero/constants.hpp#L58-L59) has units of kg/mol, which lines up with what I found with a google (28.96 g/mol == 0.02896 kg/mol). However, this is not in line with what is shown in `conversions.hpp`.
```
/// Molecular weight of dry air [kg/mol]
static constexpr Real molec_weight_dry_air = 0.028966;
```

So, a couple questions:
  - Would it be correct to assume that we prefer the **kilo-** units?
  - Does anyone have a strong preference on changing the haero molecular weight units to kg/kmol?
    - The other option is to introduce the factor of 1000 to the calculations in `conversions.hpp`.

@pressel @jeff-cohere @pbosler @odiazib

mjs271:
Haero gives `Constants::molec_weight_dry_air` in units of kg/mol, whereas `conversions.hpp` requires kg/kmol for `mmr_from_vmr()` and `vmr_from_mmr()`. So, we multiply by `constexpr Real mol_to_kmol = 1e-3` to account for this.

Closes #152 

**Note:** I realized my build was crashing due to unused variables, so I cherry-picked some commits from `collab/nucleate_ice` that fixed them.

### DIFF
--- a/src/mam4xx/conversions.hpp
+++ b/src/mam4xx/conversions.hpp
@@ -8,6 +8,7 @@
 
 #include <haero/constants.hpp>
 #include <haero/math.hpp>
+#include <mam4xx/aero_modes.hpp>
 
 /// This file contains functions for converting between various representations
 /// of physical quantities in aerosol parameterizations.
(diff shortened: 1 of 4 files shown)


pressel (approved):
Looks good to me.

odiazib (approved):

src/mam4xx/conversions.hpp
```diff
@@ -19,6 +19,8 @@ using Constants = haero::Constants;
 using haero::cube;
 using haero::square;
 
+constexpr Real mol_to_kmol = 1e-3;
+
 /// Given a number concentration for a species or mixture [m-3], computes and
```
pbosler, line 24: I could be wrong, but I think we have a few terms mixing in here, and we have to be certain of each one's units.

"stuff" can be number, volume, or mass
- mixing ratio : amount of stuff per kg dry air
- concentration: volume of stuff per m3 dry air

A volume mixing ratio is not a number concentration; vmr is m3 aerosol per kg dry air, so at minimum this comment needs to be changed.  Perhaps the function name, too; perhaps the function's content.

src/mam4xx/conversions.hpp
```diff
@@ -56,7 +58,9 @@ KOKKOS_INLINE_FUNCTION Real number_conc_from_mmr(Real mmr, Real molecular_wt,
 /// @param [in] molecular_wt The molecular weight of the species/mixture
 /// [kg/kmol]
 KOKKOS_INLINE_FUNCTION Real mmr_from_vmr(Real vmr, Real molecular_wt) {
-  const auto mw_dry_air = Constants::molec_weight_dry_air;
+  // we convert here since haero::Constants uses SI units.
```
pbosler, line 61: I thought these functions would use densities, not molecular weights:

vmr [m3 aero / kg dry air] *  (density aerosol [kg aerosol / m3 aerosol]) = mmr [kg aerosol / kg dry air]

pbosler (changes requested):

mjs271:
> I'm not in favor of putting non-SI units in generic conversion functions, such as the ones in `conversions.hpp`. I think it's best to keep `conversions.hpp` all with respect to SI units, e.g., conversion from one SI unit to another.

Makes total sense to me, in terms of changing the first 4 functions in `conversions.hpp` to use SI units (and proper formulation), namely changing `kmol -> mol`. It appears that `rename()` is the only process currently using any of these conversions (`aero_model()` is likely to also do so), so that will be relatively low-impact.

>  Then, if a specific process needs a different unit, that's where the `mol_to_kmol` conversion should go.

Works for me!

src/mam4xx/conversions.hpp
mjs271: I have no reason not to accept the definitions you give here, since I've only seen these terms used in the context of my work here and accepted the calculations in these functions as correct (oops). I will rewrite the functions in terms of those corrected quantities and tag you for a once-over when I commit.

src/mam4xx/conversions.hpp
mjs271: Given the definitions above, that makes perfect sense to me--will change!

PR #156, opened 2023-04-06
Refactor mixing ratio conversion functions

### LINKED ISSUE #152: Issue with units in vmr $\leftrightarrow$ mmr conversion calculations

mjs271:
This PR is the older sibling of #155 and the result of @pbosler's comments therein.

Based on Pete's feedback, I renamed/changed/corrected a few functions and also added a couple new ones. These functions are:
- `vol_mr_to_mass_mr()`
- `mass_mr_to_vol_mr()`
- `vol_mr_to_molar_mr()`
- `molar_mr_to_vol_mr()`
- `mass_mr_to_molar_mr()` (**_New_**: this is what was actually intended to be used in `rename()`)
- `molar_mr_to_mass_mr()` (**_New_**)

In the current form, these functions do relatively basic arithmetic calculations, perhaps begging the question of whether they should even be included. I suppose it saves some ugly indexing or extra function calls in the main source files, but I'm fine if others would prefer to get rid of some/all.

I would also ask others to take a look at the convention I use of passing the aerosol's array index as an argument. It seemed to be an efficient way of grabbing density and molecular weight. Also seemed convenient since aerosols are often getting looped over when these functions are called, meaning their index is close at hand. However, it does rely on the calling process to be sure it isn't passing an invalid index (i.e., one not included in the mode of interest or a number higher than 6).

@pressel @odiazib @singhbalwinder @overfelt @jeff-cohere (Jeff, please don't read this until next week 🙂)

Closes #152

### DIFF
--- a/src/mam4xx/conversions.hpp
+++ b/src/mam4xx/conversions.hpp
@@ -92,7 +92,7 @@ KOKKOS_INLINE_FUNCTION Real vol_mr_to_molar_mr(Real vol_mr, int aero_idx) {
 /// @param [in] aero_idx The array index for the aerosol of interest
 KOKKOS_INLINE_FUNCTION Real molar_mr_to_vol_mr(Real molar_mr, int aero_idx) {
   AeroSpecies spec = mam4::aero_species(aero_idx);
-  return molar_mr / spec.density * spec.molecular_weight;
+  return molar_mr / (spec.density * spec.molecular_weight);
 }
 
 /// Given a mass mixing ratio (mass_mr) for a species or mixture
--- a/src/mam4xx/conversions.hpp
+++ b/src/mam4xx/conversions.hpp
@@ -8,6 +8,7 @@
 
 #include <haero/constants.hpp>
 #include <haero/math.hpp>
+#include <mam4xx/aero_modes.hpp>
 
 /// This file contains functions for converting between various representations
 /// of physical quantities in aerosol parameterizations.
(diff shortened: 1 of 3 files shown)


jeff-cohere:
For the sake of consideration, what do folks think about naming the functions `x_from_y` instead of `y_to_x`? This assigns a noun `x` to the return value of the function and indicates what it's calculated from (`y`).

mjs271:
Haha, that was a rogue decision by me because that's the way it made sense in my head. As in, this is the gadget that transforms `y` into `x`. However, the `from` syntax does make a lot of sense, since the return value is the first thing mentioned.

In the end, it doesn't make any difference to me and am happy to swap back.

mjs271:
Closing for further consideration of the correct path forward.

PR #158, opened 2023-04-11
Collab/aero model

odiazib:
This PR contains the C++ port of three Fortran subroutines from `aero_model.hpp`:
[`modal_aero_bcscavcoef_init`]( https://github.com/eagles-project/e3sm_mam4_refactor/blob/b888b0f0847d870d89af9a450a0dfce6b31dec8d/components/eam/src/chemistry/modal_aero/aero_model.F90#L1637), [`modal_aero_bcscavcoef_get`]( https://github.com/eagles-project/e3sm_mam4_refactor/blob/b888b0f0847d870d89af9a450a0dfce6b31dec8d/components/eam/src/chemistry/modal_aero/aero_model.F90#L1709), and [`calc_1_impact_rate`]( https://github.com/eagles-project/e3sm_mam4_refactor/blob/b888b0f0847d870d89af9a450a0dfce6b31dec8d/components/eam/src/chemistry/modal_aero/aero_model.F90#L1784). In addition to these subroutines, we ported the following subroutines: `air_dynamic_viscosity`, `air_kinematic_viscosity`, `calc_rain_drop_conc`, `calc_aer_conc_frac`, `calc_schmidt_number`, `calc_impact_efficiency`.
We validated the C++ port using three validation tests from e3sm. All tests passed with threshold error of 2e-4. Using the same constant values from e3sm will results in smaller threshold error; however, we decided to use the constant values from `haero::Constants`.

Comments about this port:

1. The subroutine `modal_aero_bcscavcoef_get` has two loops (`icol` and `kk`). Its implementation in C++ does not have these loops. As a result, inputs and outputs from this function are local at `kk` and `icol`.
2. In `modal_aero_bcscavcoef_init` , `scavimptblvol` and `scavimptblnum` (according to comments in the fortran code are the scavenging removal in vol and number base) have a non-standard index notation $\in$ [-7,12]. In mam4xx, we used the standard index notation in C++, i.e., $\in$ [0,19]. We updated the `modal_aero_bcscavcoef_init` function to adapt the new C++ index notation, and we validated this subroutine with E3SM validation tests.
3. Global variables and public variables in the Fortran version of `modal_aero_bcscavcoef_init` are defined as inputs in C++ port version. For instance, `dgnum_amode` and `sigmag_amode`.
4. `modal_aero_bcscavcoef_init` use two arrays `specdens_amode` and `lspectype_amode` to get the dry aerosol density. We simplified and used dry aero density as an input.
5. We noted that in `modal_aero_bcscavcoef_init`, the dry aerosol density is set as the density of the first species in each mode. In this case, sulfate (1770kg/m3), sulfate (1770kg/m3), dust (2600kg/m3), and pom (1000kg/m3). We mentioned this assumption to the Fortran team.


src/mam4xx/aero_model.hpp
```diff
@@ -0,0 +1,626 @@
...
+  const Real precip = one / Real(36000.); //  1 mm/hr in cm/s
+
+  // set the iteration radius for rain droplet
+  // rain droplet bin information [cm]
+  const Real rlo = .005;
+  const Real rhi = .250;
+  const Real dr = 0.005;
+  // Nearest whole number: nint
+  // number of rain bins
+  const int nr = Real(1) + int((rhi - rlo) / dr);
```
mjs271, line 424: would `int(1) + round((rhi - rlo) / dr)` make more sense here, since the fortran code is using `nint()` (nearest integer)?


Alternatively: `const in nr = 50;` 🥸

src/mam4xx/aero_model.hpp
```diff
@@ -0,0 +1,626 @@
...
+
+  if (nr > nrainsvmax) {
+    printf("subr. calc_1_impact_rate -- nr > nrainsvmax \n ");
+    return;
+  }
```
mjs271, line 429: `if (50 > 50)`? This descent into madness continues...

src/mam4xx/aero_model.hpp
```diff
@@ -0,0 +1,626 @@
...
+  // standard deviation (log-normal distribution)
+  const Real sx = haero::log(sigmag);
+  // log(mean radius) (log-normal distribution)
+  const Real xg0 = haero::log(ag0);
+  const Real xg3 = xg0 + Real(3.) * sx * sx; // mean + 3*std^2
+
+  // set the iteration radius for aerosol particles
+  const Real dx = haero::max(0.2 * sx, 0.01);
+  const Real xlo = xg3 - haero::max(4. * sx, 2. * dx);
+  const Real xhi = xg3 + haero::max(4. * sx, 2. * dx);
```
mjs271, line 443: Should we call these `// BAD CONSTANT`s?

mjs271 (approved):
Just a few small things and no dealbreakers. I say send the real PR after a few fixups

src/mam4xx/aero_model.hpp
```diff
@@ -0,0 +1,626 @@
...
+  // @param [out]  scavimptblnum scavenging rate of aerosol number [1/s]
+  // @param [out]  scavimptblnum scavenging rate of aerosol volume [1/s]
+  // @aerosol_dry_density [in] aerosol dry density [k/m3]
+  // FIXME : create an 4 elements array that contains the aerosol densities to
+  // replace specdens_amode and lspectype_amode
+
+  const Real zero = 0;
+  const Real one = 1;
+  // ! set up temperature-pressure pair to compute impaction scavenging rates
+  // BAD CONSTANT
```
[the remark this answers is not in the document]
odiazib: Yes, it is.

src/mam4xx/aero_model.hpp
```diff
@@ -0,0 +1,626 @@
...
+  const Real press_750hPa = 0.75e6;                      //  ! dynes/cm2
+  for (int imode = 0; imode < AeroConfig::num_modes(); ++imode) {
+    const Real sigmag = sigmag_amode[imode];
+    // Note: we replaced lspectype_amode and lspectype_amode for
+    // dry_aero_density const int ll = lspectype_amode[0][imode]; const Real
+    // rhodryaero = specdens_amode[ll];
+    const Real rhodryaero = aerosol_dry_density[imode];
+    for (int jgrow = nimptblgrow_mind; jgrow <= nimptblgrow_maxd; ++jgrow) {
+      // ratio of diameter for wet/dry aerosols [fraction]
+      const Real wetdiaratio = haero::exp(jgrow * dlndg_nimptblgrow);
```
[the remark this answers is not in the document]
odiazib: It could matter if we have a division with an integer. In this case dlndg_nimptblgrow is real. However, I convert `jgrow` to real inside the `exp` function.

src/mam4xx/aero_model.hpp
```diff
@@ -0,0 +1,626 @@
...
+  const Real one = 1;
+  if (isprx_kk) {
+    // ! interpolate table values using log of
+    // (actual-wet-size)/(base-dry-size) ratio of wet and dry aerosol diameter
+    // [fraction]
+    const Real wetdiaratio = dgn_awet_imode_kk / dgnum_amode_imode;
+    // BAD CONSTANT
+    Real scavimpvol, scavimpnum = zero;
+    if (wetdiaratio >= 0.99 && wetdiaratio <= 1.01) {
+      scavimpvol = scavimptblvol[-nimptblgrow_mind][imode];
```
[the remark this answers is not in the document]
odiazib: Indexing for `scavimptblnum` in C++ $\in$ [0,19]. Indexing for scavimptblnum in Fortran $\in$ [-7,12]. Hence, for this array, index 0 in Fortran will be 7 (`-nimptblgrow_mind`) in C+++. In both cases, (0 or 7) will be the 8th element in scavimptblnum.
Here I am assuming that `nimptblgrow_mind` is negative.

src/mam4xx/aero_model.hpp
odiazib: Good point.

src/mam4xx/aero_model.hpp
odiazib: I will comment out this if statement.

src/mam4xx/mam4.hpp
```diff
@@ -10,6 +10,7 @@
 
 #include <haero/aero_process.hpp>
 #include <mam4xx/aero_config.hpp>
+#include <mam4xx/aero_model.hpp>
```
jeff-cohere, line 13: We should probably make sure this header gets installed when we type `make install` (in `src/mam4xx/CMakeLists.txt`).

src/validation/aero_model/modal_aero_bcscavcoef_get.cpp
```diff
@@ -0,0 +1,68 @@
...
+void modal_aero_bcscavcoef_get(Ensemble *ensemble) {
+  ensemble->process([=](const Input &input, Output &output) {
+    const Real zero = 0;
+
+    const int ncol = int(input.get_array("ncol")[0]);
+
+    auto scavimptblvol_vector = input.get_array("scavimptblvol");
+    auto scavimptblnum_vector = input.get_array("scavimptblnum");
+
+    Real scavimptblvol[aero_model::nimptblgrow_total][AeroConfig::num_modes()] =
```
jeff-cohere, line 24: You might not need the `zero` here--I think C++ will zero-initialize a data structure with just the curly braces. Best to verify, though.

src/validation/aero_model/modal_aero_bcscavcoef_init.cpp
```diff
@@ -0,0 +1,50 @@
+// mam4xx: Copyright (c) 2022,
```
jeff-cohere, line 1: My comments for `modal_aero_bcscavcoef_get.cpp` also apply to this file, but they're minor, so don't give them too much thought. :-)

src/mam4xx/aero_model.hpp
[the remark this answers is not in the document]
jeff-cohere: I didn't know you could use simple Latex math expressions in GitHub!

@odiazib , it may be worth adding a comment describing the indexing scheme as you have above, just so people aren't confused later on.

src/mam4xx/aero_model.hpp
```diff
@@ -0,0 +1,626 @@
...
+      }
+
+      if (jgrow < nimptblgrow_mind) {
+        jgrow = nimptblgrow_mind;
+        xgrow = jgrow;
+      } else {
+        jgrow = haero::min(jgrow, nimptblgrow_maxd - 1);
+      }
+      const Real dumfhi = xgrow - jgrow;
+      const Real dumflo = one - dumfhi;
```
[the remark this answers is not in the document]
jeff-cohere: If either or both of you understand what's happening here, it would be nice to add some comments.

src/mam4xx/aero_model.hpp
```diff
@@ -0,0 +1,634 @@
...
+Real air_dynamic_viscosity(const Real temp) {
+  /*-----------------------------------------------------------------
+  ! Calculate dynamic viscosity of air, unit [g/cm/s]
+  !
+  ! note that this calculation is different with that used in dry deposition
+  ! see the same-name function in modal_aero_drydep.F90
+  !-----------------------------------------------------------------*/
+  // @param [in] temp   ! air temperature [K]
+  // @return     dynamic viscosity of air, unit [g/cm/s]
+  return 1.8325e-4 * (416.16 / (temp + 120.)) * haero::pow(temp / 296.16, 1.5);
```
jeff-cohere, line 112: Can we provide a citation for this expression?

src/mam4xx/aero_model.hpp
```diff
@@ -0,0 +1,634 @@
...
+    } else {
+      vfallstp = 1069.8 * haero::pow(dd, 0.235);
+    }
+    // rain droplet falling speed [cm/s]
+    vfallrainsv[ii] = vfallstp * haero::sqrt(1.204e-3 / rhoair);
+    // sum of precipitation in all bins
+    precipsum += vfallrainsv[ii] * rr * rr * rr * xnumrainsv[ii];
+
+  } // ii
+    // 1.333333 is simplified 4/3 for sphere volume calculation
```
jeff-cohere, line 177: If you add a space above this line, `clang-format` will probably do a better job of indenting this comment.

src/mam4xx/aero_model.hpp
```diff
@@ -0,0 +1,634 @@
...
+
+  // @param [out]  schmidt       ! Schmidt number [unitless]
+  // @param [out]  taurelax      ! relaxation time for Stokes number
+  // [s]
+
+  // BAD CONSTANT
+  const Real boltz =
+      haero::Constants::boltzmann; // 1.38065e-23;      //  ! Boltzmann's
+                                   // constant ~ J/K/molecule
+  // BAD CONSTANT
```
jeff-cohere, line 267: This looks to me like a straight conversion from SI (mks, Joules/K) to cgs (ergs/K) units, which explains the conversion factor. Do you agree?

src/mam4xx/aero_model.hpp
```diff
@@ -0,0 +1,634 @@
...
+
+  const Real zero = 0;
+  const Real one = 1;
+  const Real three = 3;
+  // BAD CONSTANT
+  const Real dlndg_nimptblgrow = haero::log(1.25);
+  // ! set up temperature-pressure pair to compute impaction scavenging rates
+  const Real temp_0C = haero::Constants::melting_pt_h2o; //     ! K
+  // BAD CONSTANT
+  const Real press_750hPa = 0.75e6; //  ! dynes/cm2
```
jeff-cohere, line 559: This seems to be storing the pressure p = 750 hPa expressed in dynes/cm2. In that sense it's just a reference pressure and not necessarily a `BAD CONSTANT`.

src/mam4xx/aero_model.hpp
```diff
@@ -0,0 +1,634 @@
...
+    } // jgrow
+
+  } // end imode
+
+} // modal_aero_bcscavcoef_init
+
+} // end namespace aero_model
+
+/// @class aero_model
+/// This class implements MAM4's aero_model parameterization.
```
jeff-cohere, line 613: Is `aero_model` a parameterization, or is it a collection of parameterizations like the microphysics stuff that James is working on? If the latter, we might not need a class, but only a top-level function that gets called (eventually).

jeff-cohere (approved):
Looks good! I left several comments, but nothing major.

src/validation/aero_model/modal_aero_bcscavcoef_get.cpp
```diff
@@ -0,0 +1,68 @@
...
+        {{zero}};
+    Real scavimptblnum[aero_model::nimptblgrow_total][AeroConfig::num_modes()] =
+        {{zero}};
+
+    int count = 0;
+    for (int imode = 0; imode < AeroConfig::num_modes(); ++imode) {
+      for (int i = 0; i < aero_model::nimptblgrow_total; ++i) {
```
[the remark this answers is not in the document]
odiazib: It matters. These variables are saved using a row-major order; see [here](https://github.com/eagles-project/e3sm_mam4_refactor/blob/b888b0f0847d870d89af9a450a0dfce6b31dec8d/components/eam/src/chemistry/utils/yaml_input_file_io.F90#L1140). Thus, we need to use row-major order to convert a 1D vector to a 2D vector. 

I will add a comment in this test to make this point.

src/mam4xx/aero_model.hpp
```diff
@@ -0,0 +1,634 @@
...
+  const Real ag0 = dg0 / two; // mean radius of aerosol
+  // standard deviation (log-normal distribution)
+  const Real sx = haero::log(sigmag);
+  // log(mean radius) (log-normal distribution)
+  const Real xg0 = haero::log(ag0);
+  const Real xg3 = xg0 + Real(3.) * sx * sx; // mean + 3*std^2
+  // BAD CONSTANT
+  // set the iteration radius for aerosol particles
+  const Real dx = haero::max(0.2 * sx, 0.01);
+  const Real xlo = xg3 - haero::max(4. * sx, 2. * dx);
```
tangshuaiqi, line 446: It confuses me a little for the different ways of using numbers. I have seen numbers defined as `const read two = 2`, use of `Real(3.)` and direct use of` 2.` . Any specific reason to use different ways?

tangshuaiqi (approved):
looks good. just have a question on the numbers that are used in the code

src/mam4xx/aero_model.hpp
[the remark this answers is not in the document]
odiazib: maybe @pressel, @singhbalwinder, and @tangshuaiqi could help us with this.

src/mam4xx/aero_model.hpp
odiazib: I add this web page (pdf) in teh code as reference .

src/mam4xx/aero_model.hpp
odiazib: Yes, I do.

src/mam4xx/aero_model.hpp
jeff-cohere: Some of the noise comes from typecasting. For example,  `Real(3.)` seems unnecessary, because `3.` is a floating point number already, but C++ interprets constants as double precision, so maybe this expression convert it to single precision where needed. But I don't actually think most of this is necessary unless we see that it causes a problem.

In any case, I don't believe we're following consistent rules for this across the whole team, so I wouldn't worry about it too much. @odiazib or @mjs271 , if I'm oversimplifying, please chime in.

odiazib: Hi @jeff-cohere, you are correct. I like to defined number of const real, if the number appears more than twice. However, I need to be consistent with this approach.

src/mam4xx/aero_model.hpp
```diff
@@ -174,8 +183,9 @@ void calc_rain_drop_conc(const int nr, const Real rlo, const Real dr,
     precipsum += vfallrainsv[ii] * rr * rr * rr * xnumrainsv[ii];
 
   } // ii
-    // 1.333333 is simplified 4/3 for sphere volume calculation
-  precipsum *= haero::Constants::pi * Real(4) / Real(3);
+
+  // 1.333333 is simplified 4/3 for sphere volume calculation
```
jeff-cohere, line 187: This comment no longer applies now that we have `four_thirds`, right?

odiazib: Yes.

src/mam4xx/aero_model.hpp
odiazib: In aero_model, we have a subroutine that invokes microphysics.  Thus, I believe. Yes, it is a collection of parameterizations.  I will delete the class.

src/mam4xx/aero_model.hpp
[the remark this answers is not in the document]
singhbalwinder: The whole purpose of the code above is to find `jgrow`, which is an index for arrays `scavimptblvol` and `scavimptblnum`. We compute `xgrow` (a real number) and take `int` of it to get our index `jgrow`. Since `xgrow` can be negative, we change jgrow appropriately (the `int` function rounds towards zero in both Fortran and C).  

After that, we are ensuring that jgrow is within the bounds of `nimptblgrow_maxd` and `nimptblgrow_mind`. We compute `dumfhi` and `dumflo` to some sort of interpolation.

We can add comments similar to the following in the code:

```
      Real xgrow = haero::log(wetdiaratio) / dlndg_nimptblgrow;
      int jgrow = int(xgrow); // get index jgrow
      if (xgrow < zero) { // adjust jgrow appropriately if xgrow is negative
        jgrow = jgrow - 1;
      }
      // bound jgrow within max and min values
      if (jgrow < nimptblgrow_mind) {
        jgrow = nimptblgrow_mind;
        xgrow = jgrow;
      } else {
        jgrow = haero::min(jgrow, nimptblgrow_maxd - 1);
      }
     // compute factors for interpolating impaction scavenging removal amounts
      const Real dumfhi = xgrow - jgrow;
      const Real dumflo = one - dumfhi;
```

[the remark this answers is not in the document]
jeff-cohere: @odiazib or @mjs271 , can you add @singhbalwinder 's suggested comments (or something like them) to this to the code above?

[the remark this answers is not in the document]
odiazib: Done.

jeff-cohere:
@overfelt , @singhbalwinder , @pressel , any feedback on this PR?

singhbalwinder (approved):
Looks great! I have left minor clarification questions just for my understanding.

src/mam4xx/aero_model.hpp
odiazib: @singhbalwinder, here are a few comment about defining numbers.

src/mam4xx/aero_model.hpp
```diff
@@ -0,0 +1,633 @@
...
+      // bound jgrow within max and min values
+      if (jgrow < nimptblgrow_mind) {
+        jgrow = nimptblgrow_mind;
+        xgrow = jgrow;
+      } else {
+        jgrow = haero::min(jgrow, nimptblgrow_maxd - 1);
+      }
+      // compute factors for interpolating impaction scavenging removal amounts
+      const Real dumfhi = xgrow - jgrow;
+      const Real dumflo = one - dumfhi;
```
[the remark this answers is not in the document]
odiazib: @tangshuaiqi had a similar question. Here is @jeff-cohere 's reply:
> 0 and 1 are special numbers, often used to initialize things. But to me this is a matter of taste.

In addition to Jeff's comment. I also use one for 1 or zero for 0 to avoid issues between single and double precision.    

The concept of "magic number" seems to apply to constant value such as $\pi$.

src/mam4xx/aero_model.hpp
```diff
@@ -0,0 +1,633 @@
...
+  const Real four_thirds = 4. / 3.;
+  // loop over  cloud bins
+  for (int ii = 0; ii < nr; ++ii) {
+    // rain radius in the bin [cm]
+    const Real rr = rlo + ii * dr;
+    rrainsv[ii] = rr;
+    xnumrainsv[ii] = haero::exp(-rr / 2.7e-2);
+    // rain diameter in the bin [cm]
+    const Real dd = 2. * rr;
+    Real vfallstp = zero;
```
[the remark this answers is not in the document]
odiazib: We could do `Real vfallstp`; however, I like to initialize variables.

src/mam4xx/aero_model.hpp
```diff
@@ -0,0 +1,633 @@
...
+  const Real dr = 0.005;
+  // // Nearest whole number: nint
+  // // number of rain bins
+  // const int nr = 1 + haero::round((rhi - rlo) / dr);
+  // FIXME: values to compute nr are hard-coded.
+  const int nr = 50;
+  // We comment this line because nr is hard-coded.
+  // if (nr > nrainsvmax) {
+  //   printf("subr. calc_1_impact_rate -- nr > nrainsvmax \n ");
+  //   return;
```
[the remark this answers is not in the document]
odiazib: I was not sure how to hand errors in mam4xx. @jeff-cohere, could you please help?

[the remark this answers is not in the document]
jeff-cohere: This is a good question that we don't yet have an answer for. If this is indeed a runtime eeror from a failure to converge, we need to swt an error flag and a meesage to indicate failure. This calculation runs on the GPU, so we can't just throw an exception.

Can someone post a link to the Fortran code? I don't think we need a solution for this PR, but I can create an issue for this and put something together.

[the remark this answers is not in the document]
tangshuaiqi: Below is the FORTRAN code: https://github.com/eagles-project/e3sm_mam4_refactor/blob/refactor-maint-2.0/components/eam/src/chemistry/modal_aero/aero_model.F90#L1840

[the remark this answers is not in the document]
jeff-cohere: @odiazib points out [here](https://github.com/eagles-project/haero/issues/447) that another Kokkos-based code indicates runtime errors with

```
Kokkos::abort("error message \n");
```

I see no reason not to do this here. Any objections? If not, can you modify this code to call `Kokkos::abort` with an appropriate error message, @odiazib ?

src/mam4xx/aero_model.hpp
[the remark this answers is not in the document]
singhbalwinder: Initializing the variables is a good idea!

jeff-cohere:
@pressel , any objections to merging this PR?

pressel (approved):
Looks, great!  Please merge.

PR #164, opened 2023-04-28
Wet Deposition - clddiag

CameronRutherford:
This is a port of wet deposition with unit tests for some functions, and validation testing implemented for clddiag.

Reference code is in `eam/src/chemistry/aerosol/wetdep.F90`


src/mam4xx/wet_dep.hpp
```diff
@@ -0,0 +1,178 @@
...
+  Real vfall = 0.0; // Falling velocity [m/s].
+
+  // Define constant convfw from cldwat.F90
+  // Reference: Tripoli and Cotton (1980)
+  convfw = 1.94 * 2.13 * haero::sqrt(Constants::density_h2o * Constants::gravity * 2.7e-4);
+
+  for (int i = 0; i < pver; i++) {
+      rain[i] = 0.0;
+      if( temperature[i] > Constants::melting_pt_h2o ) {
+          rho = pmid[i] / (Constants::r_gas_dry_air * temperature[i]);
```
CameronRutherford, line 143: Same with these constants - I also wasn't sure if these things needed commenting/explaining as I am not a climate scientist and none of this is intuitive.

src/tests/mam4_wet_deposition_unit_tests.cpp
```diff
@@ -0,0 +1,170 @@
...
+  mam4::AeroConfig mam4_config;
+  mam4::WetDepositionProcess::ProcessConfig process_config;
+  mam4::WetDepositionProcess process(mam4_config, process_config);
+  REQUIRE(process.name() == "MAM4 Wet Deposition");
+  REQUIRE(process.aero_config() == mam4_config);
+}
+
+TEST_CASE("test_local_precip_production", "mam4_wet_deposition_process") {
+  ekat::Comm comm;
+  // TODO - figure out how to print this logging information...
```
CameronRutherford, line 38: I don't know where this sort of thing is documented. I even noticed other tests were using blanket cout statements to print instead which was odd...

src/tests/mam4_wet_deposition_unit_tests.cpp
```diff
@@ -0,0 +1,170 @@
...
+  ekat::logger::Logger<> logger("wet deposition local precip production test",
+                                ekat::logger::LogLevel::debug, comm);
+  int nlev = 72;
+  Real pblh = 1000;
```
CameronRutherford, line 42: No idea why these magic numbers, but other tests use them

src/tests/mam4_wet_deposition_unit_tests.cpp
```diff
@@ -0,0 +1,170 @@
...
+  mam4::wetdep::local_precip_production(pdel, source_term, sink_term, lprec,
+                                        atm);
+
+  // TODO - generate/use real validation data
+  for (int i = 0; i < pver; i++) {
+    REQUIRE(pdel[i] == 1.0);
+    REQUIRE(source_term[i] == 2.0);
+    REQUIRE(sink_term[i] == 1.5);
+    REQUIRE(lprec[i] ==
+            pdel[i] / Constants::gravity * (source_term[i] - sink_term[i]));
```
CameronRutherford, line 67: This is all that happens within the C++ loop, but would be good to have actual validation data :)

CameronRutherford:
I also noticed that some tests are taking substantially longer than others - is there a standard volume of test inputs for each test? Should we limit to say 10,000 points in a lattice or something similar?

jeff-cohere:
> I also noticed that some tests are taking substantially longer than others - is there a standard volume of test inputs for each test? Should we limit to say 10,000 points in a lattice or something similar?

We don't have any hard-and-fast rules about this so far, but if tests start taking a very long time, we can certainly make such a rule. I'd be surprised if our test ensembles had 10k points typically, but I haven't looked closely.

src/tests/mam4_wet_deposition_unit_tests.cpp
jeff-cohere: 72 levels is the current "standard" number of vertical levels in an atmospheric column in E3SM, so we've been using it to mock up a "typical" column.

The planetary boundary layer /height (which distinguishes between the "near-surface" and "everything else in the atmosphere") doesn't affect most calculations, so we just stick a nice round number in there.

src/mam4xx/wet_dep.hpp
```diff
@@ -0,0 +1,178 @@
...
+void rain_mix_ratio(/* cont int ncol, */ const Real *temperature, const Real *pmid,
+                             const Real *sumppr, Real *rain, const Atmosphere &atm) {
+  const int pver = atm.num_levels();
+  Real convfw = 0.0; // Falling velocity at air density = 1 kg/m3 [m/s * sqrt(rho)].
+  Real rho = 0.0; // Air density [kg/m3].
+  Real vfall = 0.0; // Falling velocity [m/s].
+
+  // Define constant convfw from cldwat.F90
+  // Reference: Tripoli and Cotton (1980)
+  convfw = 1.94 * 2.13 * haero::sqrt(Constants::density_h2o * Constants::gravity * 2.7e-4);
```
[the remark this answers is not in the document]
jeff-cohere: Not all constants exist at the Haero ("generic aerosol package") level, since different aerosol packages use different physical assumptions. In other words, we're probably not going to have a single namespace with all the physical constants in it.

src/mam4xx/wet_dep.hpp
```diff
@@ -0,0 +1,178 @@
...
+#include <haero/math.hpp>
+#include <mam4xx/aero_config.hpp>
+
+// Based on e3sm_mam4_refactor/components/eam/src/chemistry/aerosol/wetdep.F90
+namespace mam4 {
+
+namespace wetdep {
+
+/**
+ * @brief Calculate local precipitation generation rate (kg/m2/s)
```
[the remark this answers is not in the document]
jeff-cohere: That's fine. The class is there just to tie all the behavior to an instance, and we're porting subroutines, which look like functions, so I actually like this approach.

src/mam4xx/wet_dep.hpp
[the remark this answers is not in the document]
CameronRutherford: huh - so should I avoid using these constants then? Are they subject to change in the future and should be redefined at the class level?

src/mam4xx/wet_dep.hpp
CameronRutherford: Sounds good :) These constants are correct as far as I can tell so this should be fine.

.gitignore
```diff
@@ -2,3 +2,4 @@
 .haero
 build/
 haero/
+src/tests/mam4_test_config.hpp
```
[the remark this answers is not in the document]
jeff-cohere: Sure, we can change it so it gets generated in the build directory and is included appropriately. It is only a concern when testing, so it's not going to be installed.

src/tests/mam4_wet_deposition_unit_tests.cpp
jeff-cohere: Everyone has their own testing philosophy, so I'm not surprised that people are just doing Whatever. As you can see, our small, deadline-driven porting team leaves quite a few gaps in documentation and testing polish relative to larger and more established projects. :-)

What would be less odd to you? Do you think we should use `spdlog` for reporting messages in tests?

src/mam4xx/wet_dep.hpp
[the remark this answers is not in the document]
jeff-cohere: You can use any constants that exist in that namespace. If you don't find them there, you'll have to create them or use them at the mam4xx level.

src/mam4xx/wet_dep.hpp
```diff
@@ -0,0 +1,178 @@
...
+ * @brief Calculate cloudy volume which is occupied by rain or cloud water as the
+ *        max between the local cloud amount or the:
+ *          sum above of (cloud * positive precip production)      sum total precip from above
+ *            ---------------------------------------------    X    -------------------------
+ *                   sum above of ( positive precip )             sum positive precip from above
+ * 
+ * @param[in] cld Cloud faction [fraction, unitless].
+ * @param[in] lprec Local production rate of precipitation [kg/m2/s].
+ * @param[in] is_tot_cld Flag for ???
+ *      -   TODO!! - unsure what this really means.
```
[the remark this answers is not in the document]
jeff-cohere: My guess is that this is a flag that indicates whether a vertical level in an atmospheric column is considered "cloudy". But maybe @singhbalwinder can weigh in.

.gitignore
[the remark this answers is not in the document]
CameronRutherford: If it's only a concern when testing, I think this is fine. Can we just leave this in the gitignore then?

src/tests/mam4_wet_deposition_unit_tests.cpp
CameronRutherford: ekat logging is fine, I just can't remember/not sure where it is documented how to print out the logging commands that I am putting in using that logger.

jeff-cohere: I don't think we have docs for that, so if you want to make some stuff up, feel free!

.gitignore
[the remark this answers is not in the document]
jeff-cohere: Sure, that works.

src/mam4xx/wet_dep.hpp
```diff
@@ -0,0 +1,178 @@
...
+void local_precip_production(/* cont int ncol, */ const Real *pdel,
+                             const Real *source_term, const Real *sink_term,
+                             Real *lprec, const Atmosphere &atm) {
+  const int pver = atm.num_levels();
+  for (int i = 0; i < pver; i++) {
+    lprec[i] = (pdel[i] / Constants::gravity) * (source_term[i] - sink_term[i]);
+  }
+}
+
+// clang-format off
```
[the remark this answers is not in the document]
singhbalwinder: It doesn't matter but using the `k` index instead of `i` is more natural for me when thinking about the vertical levels. Maybe it is just me :-)

src/mam4xx/wet_dep.hpp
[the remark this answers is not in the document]
singhbalwinder: When `is_tot_cld` is true, this function computes the total cloud volume. Otherwise, it computes the cloud volume of either the stratiform clouds or the convective clouds (depending upon the provided cloud fraction). @tangshuaiqi please feel free to correct me or add more info.

src/mam4xx/wet_dep.hpp
```diff
@@ -0,0 +1,178 @@
...
+  // Define constant convfw from cldwat.F90
+  // Reference: Tripoli and Cotton (1980)
+  convfw = 1.94 * 2.13 * haero::sqrt(Constants::density_h2o * Constants::gravity * 2.7e-4);
+
+  for (int i = 0; i < pver; i++) {
+      rain[i] = 0.0;
+      if( temperature[i] > Constants::melting_pt_h2o ) {
+          rho = pmid[i] / (Constants::r_gas_dry_air * temperature[i]);
+          vfall = convfw / haero::sqrt(rho);
+          rain[i] = sumppr[i] / (rho * vfall);
```
singhbalwinder, line 145: Perhaps, you are still working on this function. I just wanted to mention that the original code has a limiter here ([link](https://github.com/eagles-project/e3sm_mam4_refactor/blob/e8e9c1bdef7f4f687cdaedd33930d42eff1df6e2/components/eam/src/chemistry/aerosol/wetdep.F90#L461))

singhbalwinder (commented):
Looks great! I added some minor comments.

src/tests/mam4_wet_deposition_unit_tests.cpp
singhbalwinder: Perhaps you already know but the validation data for these routines is at https://github.com/eagles-project/e3sm_mam4_refactor/tree/refactor-maint-2.0/components/eam/src/chemistry/yaml/wetdep

src/mam4xx/wet_dep.hpp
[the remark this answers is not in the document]
CameronRutherford: @jaelynlitz mentioned something about also potentially just passing in the number of levels into all of these functions instead of the entire atmosphere object - I will change these to k and also fix this in the same fix if others agree

src/mam4xx/wet_dep.hpp
```diff
@@ -0,0 +1,266 @@
...
+  auto sumppr_all = new Real[pver];
+  auto sumppr_all_cu = new Real[pver];
+  auto sumppr_all_st = new Real[pver];
+ 
+  for (int i = 0; i < pver; i++) {
+    source_term[i] = prain[i] + cmfdqr[i];
+  }
+
+  // ...and then we pass the temporary array to local_precip_production
+  // TODO - !FIXME: Possible bug: why there is no evapc in lprec calculation?
```
CameronRutherford, line 213: This was from the fortran - not sure how to make sure someone follows up on this down the road

jeff-cohere: You could create an issue for it.

CameronRutherford:
It also seems like regular unit tests that I added are failing in single precision mode. I think that means I need to change to using `Approx()` for equivalence checking, although I don't 100% understand how that function works.

cc @jeff-cohere @overfelt @pressel if any of you can help out with your previous experience?

src/validation/wetdep/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+   COPYONLY)
+
+add_test(wetdep_validation wetdep_driver
+         ${WETDEP_VALIDATION_DIR}/clddiag_input_ts_355.yaml)
+
+set(MAM4XX_WETDEP_RESULT mam4xx_clddiag_input_ts_355)
+# Just use skywalker diff since cmake compare is too specific
+add_test(NAME wetdep_compare_output
+  COMMAND ${MAM_X_VALIDATION_DIR}/tools/skywalker_diff.py ${MAM4XX_WETDEP_RESULT} ${SKYWALKER_RESULT}_ref
+)
```
CameronRutherford, line 23: Maybe this should be single precision only?

I also wanted to point this out because the mam4xx result and the skywalker ref have very different names which was annoying to debug.

src/tests/mam4_wet_deposition_unit_tests.cpp
```diff
@@ -0,0 +1,296 @@
...
+  auto sink_term_view = Kokkos::create_mirror_view(sink_term);
+  Kokkos::deep_copy(sink_term_view, sink_term);
+  auto lprec_view = Kokkos::create_mirror_view(lprec);
+  Kokkos::deep_copy(lprec_view, lprec);
+
+  for (int i = 0; i < pver; i++) {
+    REQUIRE(pdel_view(i) == 0.3395589227E+04);
+    REQUIRE(source_term_view(i) == 0.4201774770E-07);
+    REQUIRE(sink_term_view(i) == 0.7626064109E-09);
+    REQUIRE(lprec_view(i) == Approx(0.1428546071E-04));
```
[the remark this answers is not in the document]
overfelt: I did 
```
const double epsilon = std::is_same_v<Real, double> ? 0.0001 : 0.001;
const double scale = std::abs(check_X[i][j]);
REQUIRE (X[i][j] == Approx(check_X[i][j]).scale(scale).epsilon(epsilon))
```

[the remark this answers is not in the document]
CameronRutherford: Should I change to use @overfelt's suggesting of Approx with a specific scale and epsilon, or is how I have implemented it fine? I had to add this to some other unit tests as well in an effort to get the single precision tests to pass as well.

src/mam4xx/wet_dep.hpp
CameronRutherford: See #175

CameronRutherford:
Looks like tests are passing and all conversations have been resolved. Another review (or two) and I think we can merge

jeff-cohere (commented):
In general, I think we could break up these functions (which take arrays of values on different vertical levels) into functions that each operate on a single vertical level. This would allow us to do a parallel dispatch over levels, assuming that data on different levels doesn't interact. Does this seem reasonable? 

For example, instead of

```
KOKKOS_INLINE_FUNCTION
void local_precip_production(const Real *pdel, const Real *source_term, const Real *sink_term,
                                                 Real *lprec, const Atmosphere &atm) {
  const int pver = atm.num_levels();
  for (int i = 0; i < pver; i++) {
    lprec[i] = (pdel[i] / Constants::gravity) * (source_term[i] - sink_term[i]);
  }
}
```

we might use

```
KOKKOS_INLINE_FUNCTION
void local_precip_production(const Real pdel, const Real source_term, const Real sink_term,
                                                 Real &lprec) {
  lprec = (pdel / Constants::gravity) * (source_term - sink_term);
}
```

and so on. Then the top-level function (or the driver/test,if you like) would have a `parallel_for` in which the functions would be dispatched to data for each of the vertical levels.

src/tests/mam4_wet_deposition_unit_tests.cpp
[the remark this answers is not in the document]
overfelt: If the default scale and epsilon are working, then I would not override them.  In a couple of cases I had very small and very large numbers that I decided were correct even though the check was failing, so I adjusted these accordingly.

pressel:
Thanks, @CameronRutherford this looks good to me. Once you have addressed @jeff-cohere's above comment, I'm happy to approve the PR.

PR #173, opened 2023-05-23
Expand GPU partitions and fix PNNL CI

CameronRutherford:
List of GPU partitions is now given and gcc 10 is now being used on deception - see https://github.com/eagles-project/haero/pull/451 for the associated HAERO PR


CameronRutherford:
While CI is passing here, I think it would be wise to merge the haero PR first, rebuild haero here, and then merge to make sure everything is in good health

CameronRutherford:
Should be good to merge this once tests are passing

jeff-cohere (approved):
Thanks, Cameron!

PR #181, opened 2023-06-01
Add C++ version of ma_precpevap_convproc and tests.

overfelt:
First case of running into a level-dependent process that has to be evaluated from the top level to the bottom because the computation of level n depends on the results of level n-1.  To complicate it a bit more the data at each level is itself an array of data of undetermined length making it a 2D array that needs to be sub-viewed at each level. So my first use of Kokkos::subview in mam4xx.

### DIFF
--- a/src/mam4xx/convproc.hpp
+++ b/src/mam4xx/convproc.hpp
@@ -525,6 +525,7 @@ void ma_precpevap(const Real dpdry_i, const Real evapc, const Real pr_flux,
 }
 
 //=========================================================================================
+template <typename SubView>
 KOKKOS_INLINE_FUNCTION
 void ma_precpprod(const Real rprd, const Real dpdry_i,
                   const bool doconvproc_extd[ConvProc::pcnst_extd],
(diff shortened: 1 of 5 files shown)


jeff-cohere (approved):
I'm a little concerned about the use of 2D views here, though I understand it mirrors the structure of the Fortran code. If there's no difference in how each of these constituents is treated (aside from possible relationships between the different views describing a single constituent), I think it would be an interesting exercise (at some point) to try to figure out a way to break this parameterization up into constituent-specific parts and have it operate directly on data instead of passing around all these views everywhere. @overfelt , since you're closer to this than I am, let me know what you think of this idea.

Anyway, this is fine for now. Nice work navigating this stuff.

overfelt:
For this data the number of constituents in each level is run time dependent.  The values within one array are not independent, the algorithm does move values around within one array of data within one level. I've been using 80 as the length of this data because that is what the box-model uses and I need something for testing. Looking through the Fortran code it can be from 0 to something larger than 80 so the upper bound can be quite large.  We could use a large array of ColumnView but I think there would still need to be a compile-time upper bound on the length of that array or the allocation might be tricky.  2D Kokkos views do not have that problem.

jeff-cohere:
To be clear, I'm not against using 2D views in general. But this whole thing smells like it could benefit from some discussion as to what we need from it.

PR #184, opened 2023-06-07
Add convproc functions initialize_dcondt and compute_downdraft_mixing_ratio.

overfelt:


src/mam4xx/convproc.hpp
```diff
@@ -92,6 +92,8 @@ class ConvProc {
   };
   // maxd_aspectype = maximum allowable number of chemical species
   // in each aerosol mode.
+  //
+  // TODO: Why is this not 7?
```
jeff-cohere, line 96: I'm guessing this accounts for interstitial and cloudborne species...?

jeff-cohere (approved):
Looks good!

src/mam4xx/convproc.hpp
overfelt: I'm not sure.  So far for all of the arrays that are dimensioned by this value of 14, only 7 have been used. But I am not finished with all of the functions, maybe the use of 14 instead of 7 sill become apparent.

src/mam4xx/convproc.hpp
```diff
@@ -752,7 +756,192 @@ void ma_precpevap_convproc(
...
+      }
+    }
+  }
+}
+// =========================================================================================
+// TODO: compute_downdraft_mixing_ratio uses multiple levels for computation but
+// ONLY sets a SINGLE level on output for the loop over ktop to kbot.  So, it
+// should be possible to input kk and then call in parallel from ktop to kbot.
+// It is a bit confusing that the level set is kk+1 so it would be better to
+// rewrite as kkp1=>kk and kk=>kk-1 and iterate from ktop+1 to kbot inclusive.
```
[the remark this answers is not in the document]
overfelt: Starting with the bottom functions, it is sometimes not obvious how the interface might be changed to allow for parallel over levels once integrated into higher level functions.  But I am keeping this in mind and expect to eventually have Kokkos parallel loops over teams for many but maybe not all of the sub-functions.  I figure it will not be too hard to tweak working and tested C++ code at that time.

PR #190, opened 2023-06-12
Add a few convproc functions and tests.

overfelt:
@jeff-cohere ,
I would like to get this branch merged into main so I could continue to add functions without bloating an already large merge request.  I would have already merged but for a problem with a couple of Skywalker validation tests which are commented out in the CMakeLists.txt file. I'm working with Balwinder to figure out why my results differ from the e3sm Skywalker python files and my best guess at this point is that the e3sm Skywalker files need to be regenerated.  

There are quite a few functions added, some of which could be parallel over layers within a column and some not.  I've added comments at the top of the functions for how parallelizable the function is:
* Some functions apply a non-trivial algorithm (non-associative, non-binary operator) to the previous layer to determine the values for the current layer so can not be parallel over the layers.
* Some functions use data from adjacent layers but in a read-only way and only set values on the current layer. These take a layer number, kk, as input but can be called in parallel.
* Some functions only access data in the current layer even though there is an input layer number, kk, as part of the function signature. The function signature could be changed to subscript the data outside of the function.

My plan is to determine how to call these over a Kokkos team object once the higher level functions that call these lower level functions are ported.

One thing to think about is how to call functions in other processes.  There are a few functions in wetdep that are used by convproc.   These should be stand-alone functions with simple signatures that are independent of any Atmosphere, Diagnostic or Prognostic classes. Once wetdep exists should convproc call them directly or should they be factored out into a separate translation unit?

### DIFF
--- a/src/mam4xx/convproc.hpp
+++ b/src/mam4xx/convproc.hpp
@@ -1074,6 +1074,59 @@ void aer_vol_num_hygro(const Real conu[ConvProc::pcnst_extd], const Real rhoair,
   }
 }
 // ======================================================================================
+// Note: compute_wup sets wup[kk] only so can be called in parallel over the
+// vertical level index, kk.
+KOKKOS_INLINE_FUNCTION
+void compute_wup(const int iconvtype, const int kk,
+                 const Real mu_i[/* nlev+1 */],
+                 const Real cldfrac_i[/* nlev */],
+                 const Real rhoair_i[/* nlev */], const Real zmagl[/* nlev */],
+                 Real wup[/* nlev */]) {
+  // -----------------------------------------------------------------------
+  //  estimate updraft velocity (wup)
+  //  do it differently for deep and shallow convection
+  // -----------------------------------------------------------------------
+
+  // clang-format off
+  /*
+  in :: iconvtype       ! 1=deep, 2=uw shallow
+  in :: kk              ! vertical level index
+  in :: mu_i[pver+1]     ! mu at current i (note pverp dimension) [mb/s]
+  in :: cldfrac_i[pver] ! cldfrac at current icol (with adjustments) [fraction]
+  in :: rhoair_i[pver]  ! air density at current i [kg/m3]
+  in :: zmagl[pver]     ! height above surface [m]
+  inout :: wup[pver]    ! mean updraft vertical velocity at current level updraft [m/s]
+  */
+  // clang-format on
+  // pre-defined minimum updraft [m/s]
+  const Real w_min = 0.1;
+  // pre-defined peak updraft [m/s]
+  const Real w_peak = 4.0;
+
+  const int kp1 = kk + 1;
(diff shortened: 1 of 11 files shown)


jeff-cohere:
Thanks, @overfelt . Reviewing now. Regarding e.g. wetdep--I would prefer that we keep wetdep functions in their own namespace for now (`mam4::wetdep`?) and let `convproc` call them directly. At some point, we should try to untangle these things, but I think we should do that deliberately when it's appropriate, and leave them tangled in a clear way for the moment. :-)

src/mam4xx/convproc.hpp
```diff
@@ -79,16 +79,21 @@ class ConvProc {
...
     Config &operator=(const Config &) = default;
+
+    bool convproc_do_aer = true;
+    bool convproc_do_gas = false;
   };
 
   static constexpr int num_modes = AeroConfig::num_modes();
   static constexpr int num_aerosol_ids = AeroConfig::num_aerosol_ids();
+
+  // Constantants for MAM spciesi classes
```
jeff-cohere, line 90: some typos here

src/mam4xx/convproc.hpp
```diff
@@ -1127,6 +1132,480 @@ void compute_wup(const int iconvtype, const int kk,
...
+   in :: mu_p_eudp[pver]      ! = mu_i[kp1] + eudp[k] [mb/s]
+   in :: aqfrac[pcnst_extd]   ! aqueous fraction of constituent in updraft [fraction]
+   in :: icwmr[pver]    ! Convective cloud water from zm scheme [kg/kg]
+   in :: rprd[pver]     ! Convective precipitation formation rate [kg/kg/s]
+   inout :: conu[pverp][pcnst_extd]   ! mix ratio in updraft at interfaces [kg/kg]
+   inout :: dconudt_wetdep[pverp][pcnst_extd] ! d(conu)/dt by wet removal[kg/kg/s]
+  */
+  // clang-format on
+  // cutoff value of cloud water for doing updraft [kg/kg]
+  const Real clw_cut = 1.0e-6;
```
jeff-cohere, line 1560: `BAD_CONSTANT`? Or at least something we should tag to revisit?

jeff-cohere (approved):
Merge away. Left a couple of very minor comments.

src/mam4xx/convproc.hpp
overfelt: Are we marking all non-attributed (no reference to a paper or anything else) constants as BAD_CONSTANT so we can revisit at a future date?

jeff-cohere: I don't think we've been ultra-consistent about this. Certainly we mark physical constants this way. Error and "is there anything at all" thresholds less so (though given the lack of institutional knowledge, maybe we be). I'll leave this one to your judgement. The other threshold I saw in this PR is 1e-12, which to me is "really small in a numerical sense", so I'm fine not marking those.

PR #193, opened 2023-06-16
Add compute_activation_tend, compute_updraft_mixing_ratio, compute_wetdep_tend and ma_activate_convproc.

overfelt:


src/validation/convproc/compute_wup.cpp
```diff
@@ -85,15 +85,18 @@ void compute_wup(Ensemble *ensemble) {
           for (int i = 0; i < nlev; ++i)
             wup_3[i] = wup_dev[i];
 
-          convproc::compute_wup(iconvtype, kk, mu_i, cldfrac_i, rhoair_i, zmagl,
-                                wup);
+          wup[kk] =
```
jeff-cohere, line 88: 👍 I like this change.

src/mam4xx/convproc.hpp
```diff
@@ -356,16 +704,17 @@ void update_tendency_diagnostics(
 }
 // =========================================================================================
 // clang-format off
+template<class SubView, class ConstSubView>
```
jeff-cohere, line 707: Would it make sense to consider a default for the second template parameter that just `const`ifies the first one?

src/mam4xx/convproc.hpp
```diff
@@ -608,14 +957,14 @@ void ma_precpevap(const Real dpdry_i, const Real evapc, const Real pr_flux,
 }
 
 //=========================================================================================
-template <typename SubView>
+template <typename Const_SubView, typename SubView>
```
jeff-cohere, line 960: We might want to normalize the order of template parameters in these functions (e.g. so this function matches `update_tendency_final`). Then at least all the moving parts can be in similar orders. :-)

src/mam4xx/convproc.hpp
jeff-cohere: Ah, I suppose it doesn't matter if the signature is constructed to match the arguments. Though perhaps there's an opportunity to do some type checking.

src/mam4xx/convproc.hpp
```diff
@@ -867,21 +1216,21 @@ void initialize_dcondt(const bool doconvproc_extd[ConvProc::pcnst_extd],
...
-   in :: conu(pverp,pcnst_extd)   ! mix ratio in updraft at interfaces [kg/kg]
-   in :: cond(pverp,pcnst_extd)   ! mix ratio in downdraft at interfaces [kg/kg]
-   in :: dconudt_activa(pverp,pcnst_extd) ! d(conu)/dt by activation [kg/kg/s]
-   in :: dconudt_wetdep(pverp,pcnst_extd) ! d(conu)/dt by wet removal[kg/kg/s]
-   in :: dudp(pver)           ! du(i,k)*dp(i,k) at current i [mb/s]
-   in :: dddp(pver)           ! dd(i,k)*dp(i,k) at current i [mb/s]
-   in :: eudp(pver)           ! eu(i,k)*dp(i,k) at current i [mb/s]
-   in :: eddp(pver)           ! ed(i,k)*dp(i,k) at current i [mb/s]
-   out :: dcondt(pver,pcnst_extd)  ! grid-average TMR tendency for current column  [kg/kg/s]
+   in :: dpdry(nlev)              ! dp [mb]
```
jeff-cohere, line 1219: Not a big deal, but maybe using square brackets here as you've done elsewhere would be a comfort to readers.

src/mam4xx/convproc.hpp
overfelt: The View and ConstView template parameters are the difference between Kokkos::View<Real> and Kokkos::View<const Real> so it is non-trivial to derive the const from the non-const.  Some of these functions are templated just to avoid complicated declarations of Kokkos subviews of views. Since the values are assigned to an "auto" to begin with, making it a templated function means never having to write out the complicated Kokkos type.

src/mam4xx/convproc.hpp
overfelt: Yes, I can make these templated functions more consistent in the declarations.

src/mam4xx/convproc.hpp
overfelt: Done.

jeff-cohere (approved):
Looks good! There's a lot of stuff going on here. Thanks for seeing this through.

PR #195, opened 2023-07-07
bugfix--spacing in top-level CML.txt `cmake_minimum_required()`

mjs271:
So, to my surprise, when builds weren't working out properly for me, I realized that the machine I was using was running an old version of cmake, and that fact was not caught by the `cmake_minimum_required (VERSION 3.17.0)`. As it turns out, the spacing (lack of) is rather important 🙃

Working properly now.


jeff-cohere (approved):
This is nuts! Thanks for fixing it.

PR #198, opened 2023-07-17
Kylepres/drydep functions

pressel:
A work in progress port of the remaining F90 functions for aerosol dry deposition.

### DIFF
--- a/src/mam4xx/drydep.hpp
+++ b/src/mam4xx/drydep.hpp
@@ -53,6 +53,14 @@ public:
 
 namespace drydep {
 
+// ==========================================================================
+// Calculate dynamic viscosity of air, unit [kg m-1 s-1]. See RoY94 p. 102
+// ==========================================================================
+KOKKOS_INLINE_FUNCTION
+Real air_dynamic_viscosity(const Real temp) {
+  return 1.72e-5 * (haero::pow(temp / 273.0, 1.5)) * 393.0 / (temp + 120.0);
+}
+
 //==========================================================================
 // Calculate kinematic viscosity of air, unit [m2 s-1]
 //==========================================================================
--- a/src/mam4xx/drydep.hpp
+++ b/src/mam4xx/drydep.hpp
@@ -53,6 +53,29 @@ public:
 
 namespace drydep {
 
+//====================================================================
+// Calculate the Schmidt number of air [unitless], see SeP97 p.972
+//====================================================================
+KOKKOS_INLINE_FUNCTION
+Real schmidt_number(const Real temp, const Real pres, const Real radius,
+                    const Real vsc_dyn_atm, const Real vsc_knm_atm) {
+
+  //  slip correction factor [unitless]
+  const Real slp_crc = slip_correction_factor(vsc_dyn_atm, pres, temp, radius);
+
+  // Brownian diffusivity of particle [m2/s], see SeP97 p.474
+  const Real dff_aer = Constants::boltzmann * temp * slp_crc /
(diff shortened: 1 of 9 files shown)


src/mam4xx/drydep.hpp
```diff
@@ -53,6 +53,82 @@ class DryDep {
...
+KOKKOS_INLINE_FUNCTION
+Real radius_for_moment(const int moment, const Real sig_part,
+                       const Real radius_part, const Real radius_max) {
+  const Real lnsig = haero::log(sig_part);
+  return haero::min(radius_max, radius_part) *
+         haero::exp((moment - 1.5) * lnsig * lnsig);
+}
+
+//==========================================================================
+// Calculate dynamic viscosity of air, unit [kg m-1 s-1]. See RoY94 p. 102
```
jeff-cohere, line 68: Thanks for the reference. Is this abbreviation spelled out anywhere?

src/validation/drydep/radius_for_moment.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <validation.hpp>
+
+using namespace skywalker;
+using namespace mam4;
+
+void radius_for_moment(Ensemble *ensemble) {
+
+  // Run the ensemble.
+  ensemble->process([=](const Input &input, Output &output) {
+    if (!input.has("moment")) {
```
jeff-cohere, line 18: Not at all important, but repetitive code like this is ripe for a CPP macro, e.g.

```
#define CHECK_FOR_INPUT(name) \
if (!input.has(#name)) { \
  std::cerr << "Required name: " << #name << std::endl; \
  exit(1); \
}

...
  CHECK_FOR_INPUT("moment")
  CHECK_FOR_INPUT("sig_part")
  CHECK_FOR_INPUT("radius_part")
  CHECK_FOR_INPUT("radius_max")
```

But this is largely a matter of taste.

pressel: Totally, we should have done that a long time ago.  I think we need to go back and remove this repetitive code. In fact, I vaguely remember now that we have a macro defined somewhere that already does just this.

jeff-cohere: If we don't already have one, you could stick such a macro in `src/validation/validation.hpp`. And you might change the header guard at the top of the file to `MAM4XX_VALIDATION` instead of `HAERO_VALIDATION`. :-)

src/mam4xx/drydep.hpp
```diff
@@ -53,6 +53,82 @@ class DryDep {
...
+  // [m]
+  const Real mean_free_path =
+      2.0 * dyn_visc /
+      (pres *
+       haero::sqrt(8.0 / (Constants::pi * Constants::r_gas_dry_air * temp)));
+
+  const Real slip_correction_factor =
+      1.0 +
+      mean_free_path *
+          (1.257 + 0.4 * haero::exp(-1.1 * particle_radius / mean_free_path)) /
```
odiazib, line 103: BAD CONSTANT for 1.257 ?  Maybe  we need to ask for a reference.

odiazib (approved):
Looks great!

src/validation/drydep/radius_for_moment.cpp
pressel: Remove as discussed.

pressel:
Once CI finishes I will merge. Thanks for the super fast reviews @jeff-cohere and @odiazib.

PR #204, opened 2023-07-20
This tiny fix makes sure mam4xx's gas chemistry headers are installed.

jeff-cohere:
I am trying to keep our eamxx integration work up to date with the haero and mam4xx repos, and occasionally I discover that we're not installing everything we need for eamxx. :-)


odiazib (approved):
Thanks Jeff!

PR #206, opened 2023-07-24
First PR for water uptake.

pressel:
Functions: 

- water_uptake_wetdens
- find_real_solution
- makoh_quartic

Comments:
On interesting part of this PR is that we discovered that F90 and Kokkos seem to treat raising complex numbers to fractional powers differently. This lead to a difference in sign of the imaginary part of the return values from `makoh_quartic`. In the next step of the water uptake calculation, the absolute value is taken of the imaginary part of the return values so this sign should not change the end solution of the water uptake calculation. However, this sign change between Fortran and C++ necessitated that we modify our validation testing. In particular, we compare the absolute values of the imaginary part of the return values of makoh_quartic.


src/mam4xx/water_uptake.hpp
```diff
@@ -0,0 +1,162 @@
...
+}
+
+//----------------------------------------------------------------------
+//  find the smallest real solution from the polynomial solver
+//----------------------------------------------------------------------
+//
+KOKKOS_INLINE_FUNCTION
+void find_real_solution(const Real rdry, const Kokkos::complex<Real> cx[4],
+                        Real &rwet, int &nsol) {
+  rwet = 1000.0 * rdry;
```
odiazib, line 94: is this one thousand a unit factor ?

odiazib (approved):
It looks good to me.

src/mam4xx/water_uptake.hpp
pressel: Yes, I believe it must be. I'll go ahead and leave a comment to that effect. Also, it is useful to remember that `rwet` and `rdry` apparently different units coming into `find_real_solution`.

jeff-cohere (approved):
Looks great! Glad the quartic solver wasn't a big deal in the end.

pressel:
Thanks @odiazib and @jeff-cohere! I'll merge the validation data PR, update it here and then merge this PR.

PR #207, opened 2023-07-26
Second part of water uptake. 

pressel:
This PR brings in a port of the modal_aero_kohler function.

### DIFF
--- a/src/mam4xx/water_uptake.hpp
+++ b/src/mam4xx/water_uptake.hpp
@@ -181,7 +181,8 @@ void modal_aero_kohler(const Real rdry_in, const Real hygro, const Real rh,
   static constexpr Real rmax = 30.0;             // (BAD CONSTANT)
 
   // effect of organics on surface tension is neglected'
-  const Real aa = 2.0e4 * mw * surften / (ugascon * tair * rhow);
+  const Real aa =
+      2.0e4 * mw * surften / (ugascon * tair * rhow); // (BAD CONSTANT)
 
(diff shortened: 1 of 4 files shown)


src/mam4xx/water_uptake.hpp
```diff
@@ -156,6 +157,63 @@ void makoh_quartic(Kokkos::complex<Real> cx[4], const Real p3, const Real p2,
...
+  static constexpr Real ugascon = 8.3e7; // (BAD CONSTANT)
+  static constexpr Real factor_um2m =
+      1.e-6; // (BAD CONSTANT) convert micron to m
+  static constexpr Real factor_m2um =
+      1.e6; // (BAD CONSTANT) convert m to micron
+  static constexpr Real small_value_10 = 1.e-10; // (BAD CONSTANT)
+  static constexpr Real rmax = 30.0;             // (BAD CONSTANT)
+
+  // effect of organics on surface tension is neglected'
+  const Real aa = 2.0e4 * mw * surften / (ugascon * tair * rhow);
```
jeff-cohere, line 184: Any idea about this factor of 20,000?

jeff-cohere (approved):
:+1: (just a minor comment on a numerical factor). Also, it looks like you need to run `clang-format`.

src/mam4xx/water_uptake.hpp
pressel: annotated this as a bad constant

PR #209, opened 2023-07-28
Initial implementation of most of the lower level functinos.

overfelt:
This is much larger a merge request than I like, but there was somewhat of a rush to complete wetdep.  Even this merge request is not the end of wetdep. WetDeposition::compute_tendencies still needs work and is not complete.  The main task of compute_tendencies is to call wetdepa_v2 which is does.  But this call should be in a triply nested loop which is not the case as yet.  

There is also the status of aero_model.hpp.  I included aero_model.hpp in wet_dep.hpp and put some functions into aero_model that wet_dep uses.  I assume this is the correct dependency order for these two header files.  If so, then maybe aero_model is the correct place for functions and static data shared among processes.


jeff-cohere:
Does `aero_model.hpp` correspond to a Fortran module in ur-MAM4? If so, we can probably just follow the dependency structure there.

overfelt:
aero_model.hpp is in mam4.  But there it calls wetdep from a function called aero_model_wetdep.  But the aero_model_wetdep function will become compute_tendencies for WetDeposition.  Seems that most of aero_model should find it's way into the various mam4xx processes.

jeff-cohere (approved):
Looks good. It is a large PR, though, so let me know if you want me to more closely examine any particular thing.

jeff-cohere:
The non-determinism we were seeing in the PNNL CI system was caused by all threads in the team policy (in `src/validation/convproc/compute_tendencies.cpp`) simultaneously doing all the work, since there's actually no team dispatch within the function called at the top-level dispatch. Adjusting the team size to 1 fixed this issue.

PR #219, opened 2023-08-07
Implementation of wetdep::compute_tendencies().

overfelt:


src/validation/ndrop/update_from_explmix.cpp
```diff
@@ -21,8 +21,8 @@ void update_from_explmix(Ensemble *ensemble) {
     const auto mam_idx_db = input.get_array("mam_idx");
     const auto nspec_amode_db = input.get_array("nspec_amode");
 
-    int nnew_in = input.get_array("nnew")[0];
-    int nsav_in = input.get_array("nsav")[0];
+    /* int nnew_in = */ input.get_array("nnew")[0];
```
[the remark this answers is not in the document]
overfelt: I got a compile warning about the values "nnew_in" and "nsav_in" not being used, but figured the get_array would still check for existence so left them in.

src/mam4xx/aero_model.hpp
```diff
@@ -810,6 +835,171 @@ bool examine_prec_exist(const int level_for_precipitation, const Real pdel[],
...
+  // clang-format off
+  /*
+  in  :: kk;
+       state_q and ptend_q only use dust and seasalt in this subroutine
+  in  :: state_q    ! tracer of state%q [kg/kg]
+  in  :: ptend_q    ! tracer tendency (ptend%q) [kg/kg/s]
+  in  :: dt         ! time step [s]
+  out :: f_act_conv_coarse      ! prescribed coarse mode aerosol activation fraction for convective
+  out :: f_act_conv_coarse_dust ! prescribed dust aerosol activation fraction for convective cloud [fracti
+  out :: f_act_conv_coarse_nacl ! prescribed seasalt aerosol activation fraction for convective cloud [fra
```
jeff-cohere, line 858: Did the two lines above get truncated in a copy-paste?

src/mam4xx/aero_model.hpp
```diff
@@ -810,6 +835,171 @@ bool examine_prec_exist(const int level_for_precipitation, const Real pdel[],
...
+  in :: pdel       ! pressure difference between two layers [Pa]
+  in :: layer_tend ! physical tendencies in each layer [kg/kg/s]
+  out :: sflx      ! integrated surface fluxes [kg/m2/s]
+  */
+  // clang-format on
+  const Real gravit = Constants::gravity;
+
+  scratch[0] = 0;
+  team.team_barrier();
+  scratch[0] += layer_tend * pdel / gravit;
```
jeff-cohere, line 940: Is every thread accumulating a contribution to `scratch`? Is that why it's `Atomic`?

src/mam4xx/wet_dep.hpp
```diff
@@ -1335,48 +1330,228 @@ void WetDeposition::compute_tendencies(
...
+      });
+  team.team_barrier();
   ColumnView dlf = diags.total_convective_detrainment;
+
+  // I don't have a good way to integrate the values computed for each
+  // level half way through the next Kokkos::parallel_for as the function
+  // calc_sfc_flux does.
+  // TODO: Determine a better way to implement calc_sfc_flux without
+  // team_fence() to avoid locking if the team size is not nlev;
+  EKAT_KERNEL_REQUIRE(team.team_size() == nlev);
```
jeff-cohere, line 1349: Let me know if you want a sounding wall or a chat about this.

src/mam4xx/aero_model.hpp
overfelt: Yes.  This is the problem of a parallel_for over the layers of a column but needing to stop right in the middle of a triple  loop and integrate some value over the column.  Since we are already in a team parallel loop the best I could think of was for each thread to add to a Kokkos::View since the View is shared by all threads.  This requires an atomic_add.  Then a barrier() to make sure all threads are done before extracting the value out of the View so each thread can then continue.  I saw a Kokkos note that low-thread-count contention is not too bad for atomic operations and on a GPU the barrier() should be fast since all threads march in lockstep for the most part.   The other option I tried was a View of length the number of levels so each thread could put the value in the thread slot, then a barrier(), then integrate the View in each thread.   A problem with both is the barrier() which could hang if not all threads reach this point, like if the thread team is not equal to the number of levels.   I'm open to suggestions....

src/mam4xx/aero_model.hpp
overfelt: Yes, probably a cut/paste error without text wrapping turned on.

src/mam4xx/wet_dep.hpp
jeff-cohere: Have you looked into Kokkos's own [parallel_scan](https://kokkos.github.io/kokkos-core-wiki/API/core/parallel-dispatch/parallel_scan.html)?

src/mam4xx/aero_model.hpp
overfelt: It would be ugly to break up into two loops since it is so deep down and all of the local values would be lost.  Convproc is even worse since it needs to do a column integration several functions down from compute_tendencies.

src/mam4xx/wet_dep.hpp
overfelt: I think that since we are already in a team parallel loop, then Kokkos paralel_scan can not be used as we have used up the two levels of parallelism Kokkos supports.  But I could be wrong about that.   Given what we are learning about team parallel loops, maybe having the loop count match the team size would allow the parallel_scan to work. But that might be relying on unsupported behavior.

jeff-cohere: Yeah, I wouldn't think that a parallel_scan inside a parallel_for is a good idea. It sounds to me like the right answer is (eventually) to restructure this code to be more sensible in a parallel environment. But if we can get this working by stipulating that the number of threads per team is `nlev`, I think that's a reasonable short-term solution.

jeff-cohere: That may be fine. I'm not sure what role `convproc` is going to play in the final product. I suspect we'll be mining it piecemeal, in which case we'll likely be restructuring it anyway.

jeff-cohere:
Good to merge after the tests finish (if you're satisfied with our conversation)

PR #220, opened 2023-08-09
Water uptake part 4.

pressel:
This is the fourth PR associated with water uptake. It provides a port of the `modal_aero_water_uptake_dryer` subroutine.

### DIFF
--- a/src/mam4xx/water_uptake.hpp
+++ b/src/mam4xx/water_uptake.hpp
@@ -387,45 +387,45 @@ void modal_aero_wateruptake_dryaer(
 
       int la, lc;
       convproc::assign_la_lc(imode, ispec, la, lc);
-
       const Real raer = state_q[la];
       const Real vol_tmp = raer / specdens;
       maer = maer + raer;
+      dryvolmr += vol_tmp;
 
       // hygro currently is sum(hygro * volume) of each species,
       // need to divided by sum(volume) later to get mean hygro for all species
       hygro[imode] += vol_tmp * spechygro_i;
+    } // end loop over species
+    if (dryvolmr > small_value_30) {
+      hygro[imode] = hygro[imode] / dryvolmr;
+    } else {
 
-      if (dryvolmr > small_value_30) {
-        hygro[imode] /= dryvolmr;
-      } else {
-        hygro[imode] = spechygro_1;
-      }
-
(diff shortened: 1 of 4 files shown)


src/mam4xx/water_uptake.hpp
```diff
@@ -301,6 +306,129 @@ void modal_aero_water_uptake_rh_clearair(const Real temperature,
...
+
+  const int ntot_amode = AeroConfig::num_modes();
+  int nspec_amode_temp[ntot_amode] = {7, 4, 7, 3};
+
+  for (int i = 0; i < ntot_amode; ++i) {
+    nspec_amode[i] = nspec_amode_temp[i];
+  }
+
+  Real specdens_amode_temp[maxd_aspectype] = {
+      0.1770000000E+04, 0.1797693135 + 309, 0.1797693135 + 309,
```
jeff-cohere, line 323: The second and third values here look fishy to me (and also those on line 329), but they're the same, so it must be intentional(?).

src/mam4xx/water_uptake.hpp
```diff
@@ -58,7 +59,11 @@ class Water_Uptake {
 };
 
 namespace water_uptake {
-
+//  number of variables in state_q
+constexpr int nvars = 40;
+constexpr int maxd_aspectype = 14;
```
jeff-cohere, line 64: is this the number of distinct interstitial + cloudeborne species indices?

jeff-cohere (approved):
Looks good.

src/mam4xx/water_uptake.hpp
pressel: Yeah, they looked fishy to me too but are values output directly by e3sm, see[ here](https://github.com/eagles-project/e3sm_mam4_refactor/blob/c658ed322a9cc2abb9fe6dce85d45cebe872ac00/components/eam/src/chemistry/yaml/cam_ndrop/dropmixnuc_input_ts_1400.yaml).

@odiazib

src/mam4xx/water_uptake.hpp
pressel: According to comments in the F90 code, it is the number of  vapor and aerosol number/mass mixing ratio variables in state%q [#/kg or kg/kg].

PR #221, opened 2023-08-11
Oscar/mo photo

odiazib:
This PR contains the port to C++/Kokkos of the following Fortran subroutines:
[cloud_mod](https://github.com/eagles-project/e3sm_mam4_refactor/blob/86dcfc5f7162835c0221b3014b7822723360e678/components/eam/src/chemistry/mozart/mo_photo.F90#L510) , [find_index](https://github.com/eagles-project/e3sm_mam4_refactor/blob/86dcfc5f7162835c0221b3014b7822723360e678/components/eam/src/chemistry/mozart/mo_jlong.F90#L823), [calc_sum_wght](https://github.com/eagles-project/e3sm_mam4_refactor/blob/86dcfc5f7162835c0221b3014b7822723360e678/components/eam/src/chemistry/mozart/mo_jlong.F90#L773), [interpolate_rsf]( https://github.com/eagles-project/e3sm_mam4_refactor/blob/86dcfc5f7162835c0221b3014b7822723360e678/components/eam/src/chemistry/mozart/mo_jlong.F90#L637), [jlong]( https://github.com/eagles-project/e3sm_mam4_refactor/blob/86dcfc5f7162835c0221b3014b7822723360e678/components/eam/src/chemistry/mozart/mo_jlong.F90#L513)

I created validation tests using data generated by @tangshuaiqi and make sure that all tests match e3sm reference files with relative error smaller than 1e-6. The functions `interpolate_rsf`, `jlong`, and `calc_sum_wght` depend on the pre-computed table `rsf_table`, a table with 5 dimensions and shape of [67, 135, 100, 100, 100] that requires > 500Mb. Based on recommendations from @jeff-cohere, I decided to create a smaller version of this table. The smaller version of the table allows to test the ported code in a fraction of time. Note that using rsf_table with original shape, a test takes 10 mins to complete.

I generated reference files using the smaller version of rsf_table with the Fortran standalone drivers ([driver_interpolate_rsf](https://github.com/eagles-project/mam4_standalone_validation/blob/fb32d3d4964dd5edd7fee593e0e2e6fb08d284a3/src/examples/driver_interpolate_rsf.F90#L1)) and yaml inputs generated by @tangshuaiqi.


src/mam4xx/mo_photo.hpp
```diff
@@ -0,0 +1,552 @@
...
+   data which lists coss sections for temperatures from
+   150 to 350 degrees K.  Make sure the index is a value
+   between 1 and 201.
+  ------------------------------------------------------------------------------*/
+
+  for (int kk = 0; kk < nlev; kk++) {
+    /*----------------------------------------------------------------------
+      ... get index into xsqy
+     ----------------------------------------------------------------------*/
+    // BAD CONSTANT
```
jeff-cohere, line 489: Is the BAD CONSTANT below 148.5?

src/mam4xx/mo_photo.hpp
```diff
@@ -0,0 +1,552 @@
...
+#include <haero/math.hpp>
+#include <mam4xx/mam4_types.hpp>
+#include <mam4xx/utils.hpp>
+
+namespace mam4 {
+
+namespace mo_photo {
+
+// number of vertical levels
+constexpr int pver = 72;
```
jeff-cohere, line 13: @pressel and I were discussing the fact that we seem to be hard-wiring the number of vertical levels in given configurations. To be honest, this isn't a huge deal right now as long as we make it possible to configure this parameter for mam4xx at compile time, because it's also [hardwired in EAMxx](https://github.com/eagles-project/scream/blob/master/components/eamxx/src/scream_config.h.in#L20). But it's something that we should be aware of.

To make sure we don't forget about this, I'll create two issues:

1. To provide a CMake parameter for the number of vertical levels so that we can configure it in the same way as it is in EAMxx (for the short term)
2. A reminder for us to revisit this and see how we might (if practical) convert this to a run-time parameter.

jeff-cohere (approved):
Looks good, @odiazib . I added a couple of minor comments.

src/mam4xx/mo_photo.hpp
jeff-cohere: See #223 and #224

src/mam4xx/mo_photo.hpp
odiazib: Yes, It does. But, I will add another BAD CONSTANT for 201.

pressel (approved):
This looks great to me thanks @odiazib!

PR #225, opened 2023-08-14
Adds NUM_VERTICAL_LEVELS as a CMake parameter that defines mam4::nlev.

### LINKED ISSUE #223: Add CMake parameter for number of vertical levels
Some of our parameterizations fix the number of vertical levels in a column at compile time because of the way they use automatic arrays for column data. This isn't ideal, but it is okay in the short term because [EAMxx also fixes its number of vertical levels](https://github.com/eagles-project/scream/blob/master/components/eamxx/src/scream_config.h.in#L20) for a given build.

In order to properly support a build-time configurable number of vertical levels for `ColumnView`s, we need a CMake parameter that defines this number in a generated header file somewhere. It looks like we do all such header generation in Haero at the moment, but this is a mam4xx-specific issue, so let's do it in this repository.

jeff-cohere:
Because of the way we use automatic C++ arrays in some of our parameterizations, it is necessary for us to have the number of vertical levels as a compile-time constant. While this is undesirable in an ideal situation, it's actually fine in the short term because EAMxx also has a compile-time-configured number of vertical levels.

mam4::nlev now contains the number of vertical levels used internally within mam4xx's C++ parameterizations.  I haven't changed any of the test or validation code to set the number of vertical levels to mam4::nlev, since that would imply that we have datasets for other numbers of vertical levels. To make it possible for us to run validations tests in situations where mam4::nlev != 72, I have added the constraint that mam4::nlev >= 72. As long as we are zeroing our views before using them, that should allow these tests to run and pass.

**NOTE: This means that `aero_config.hpp` is now generated from `aero_config.hpp.in`.**

I've also adjusted our CI workflow scripts to explicitly set the number of vertical levels to 72. Strictly speaking, this isn't necessary, because 72 is the default value for this parameter. But I think it helps to have these things made explicit.

Closes #223

### DIFF
--- a/.github/pnnl-ci/ci.sh
+++ b/.github/pnnl-ci/ci.sh
@@ -55,6 +55,7 @@ git submodule update --init || exit
 
 cmake \
   -DMAM4XX_HAERO_DIR=$HAERO_INSTALL \
+  -DNUM_VERTICAL_LEVELS=72 \
   -DCMAKE_INSTALL_PREFIX=$(pwd)/install \
   -DCMAKE_BUILD_TYPE=$BUILD_TYPE \
   -DCMAKE_C_COMPILER=$CC \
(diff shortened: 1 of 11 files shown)


odiazib (approved):

pressel (approved):
Looks good. Thanks for acting on this so quickly @jeff-cohere!

PR #229, opened 2023-08-22
Add Comments and Concerns to gas_chem, plus one minor fix

mjs271:
In an effort to create a case in which the Newton-Raphson solve in `gas_chem.hpp` fails, I ended up having to go through most of that file and also `gas_chem_mechanism.hpp` to get things straight in my own head. I added a bunch of explanatory comments and also some spots where computational weirdness appeared to be ocurring.

In doing that, I noticed that, from a perspective of units consistency, there are some very strange-looking arithmetic calculations being done. I put a `FIXME` at the ones I spotted along with some salient details. Note that, it's ***possible*** that everything is ok, but that would seem to require the `base_solution`, `solution`, etc., arrays to be carrying quantities with differing units and/or the same for the reaction rates (`lrxt`, `rxt`) or washout rates(?) (`lhet`, `het_rates`) arrays. This would be in conflict with comments from MAM4 that define:
```
lrxt := rxt rates [1/cm^3/s]
lhet := washout rates [1/s]
solution := species mixing ratios [vmr]
```

Other than that, fixed what appears to be [one tiny transcription error](https://github.com/eagles-project/mam4xx/blob/d2412e536cdb444be51f15ae4a014b4e2b1186d4/src/mam4xx/gas_chem.hpp#L357). I would be surprised if that resulted in failing tests, but I suppose we'll see...

### DIFF
--- a/src/mam4xx/gas_chem.hpp
+++ b/src/mam4xx/gas_chem.hpp
@@ -167,7 +167,7 @@ void newton_raphson_iter(const Real dti, const Real lin_jac[nzcnt],
         converged[kk] = true;
 
         int mm = permute_4[kk];
-        // FIXME: is there a computational reason this needs to happen?
+        // TODO: is there a computational reason this needs to happen?
         // I suspect not, given that epsilon is hard-coded to 1e-3, meaning that
         // all of this logic surrounding 'converged[kk] = ...' is unnecessary
         bool frc_mask = haero::abs(forcing[mm]) > small;
@@ -274,7 +274,7 @@ void imp_sol(Real base_sol[gas_pcnst], // inout - species mixing ratios [vmr]
     //  ... set the iteration invariant part of the function f(y)
     // -----------------------------------------------------------------------
 
-    // FIXME: the units seem wrong here--could these arrays hold quantities
+    // TODO: the units seem wrong here--could these arrays hold quantities
     // with different units?
     // ind_prd has units [1/cm^3/s] (for the entries that are nonzero)
     // dti units are [1/s], and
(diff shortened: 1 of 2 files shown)


odiazib:
We need to consider that constants in gas chemistry have units. These constants do not have simple units, they depend on the order of the reaction.

mjs271:
That makes sense. I clearly just put my head down and took the old comments at face value. I suppose in the mean time, we give MAM4 the benefit of the doubt and assume things are correct.

I'll nix the comments tomorrow--thanks Oscar!

singhbalwinder:
Thanks @mjs271 for digging into these issues! Tagging @tangshuaiqi, @brigaudet, and @mingxuanwupnnl as they might clarify some of these unit mismatches.

mjs271:
After talking this over with Oscar and Jeff, I've changed the added `FIXME`s to `TODO`s and will make a record of this in a separate issue.

I do agree with @odiazib and @brigaudet that it's likely all of the units do check out. However, we eventually need to reconcile the comments with the math operations occurring in the code, but that's a problem for the future versions of ourselves (Hi! 👋), once porting work is finished. I would also argue that if certain elements in these reaction-rate or species qoi (vmr, mmr, number conc) arrays have different units from the rest of the elements, they should be in a separate array at least until the point that the same operation is being performed on the disparate quantities--e.g., the final Newton solve.

So, if I don't hear anything in the next day or so, I will merge this in.

odiazib (approved):

jeff-cohere (approved):

singhbalwinder (approved):

PR #230, opened 2023-08-22
Oscar/mo photo 2

odiazib:
This PR contains port of the [`table_photo`]( https://github.com/eagles-project/e3sm_mam4_refactor/blob/a02a973c9154fad10cc85a62f76feb03d2d62589/components/eam/src/chemistry/mozart/mo_photo.F90#L399) subroutine.

Notes:
1.	Fortran version of `table_photo` computes the photo variable for all columns; mam4xx version of `table_photo` only computes one column. 
2.	Validation reference files were generated by [driver_table_photo](https://github.com/eagles-project/mam4_standalone_validation/blob/main/src/examples/driver_table_photo.F90). I did not use reference files from e3sm because the rsf_table requires  > 500Mb; see [PR #221]( https://github.com/eagles-project/mam4xx/pull/221). However, part of the input file for `table_photo` contains inputs from e3sm.

### DIFF
--- a/src/validation/validation.hpp
+++ b/src/validation/validation.hpp
@@ -125,12 +125,11 @@ void create_synthetic_rsf_tab(View5D &rsf_tab, const int nw, const int nump,
                               const int numsza, const int numcolo3,
                               const int numalb, Real *synthetic_values);
 
-void convert_1d_std_to_2d_view_device(const std::vector<Real>& pmid_db,
-                                 const View2D& var_device);
+void convert_1d_std_to_2d_view_device(const std::vector<Real> &pmid_db,
+                                      const View2D &var_device);
 
-void convert_2d_view_device_to_1d_std(const View2D& var_device,
-                                      std::vector<Real>& var_std
-                                 );
+void convert_2d_view_device_to_1d_std(const View2D &var_device,
+                                      std::vector<Real> &var_std);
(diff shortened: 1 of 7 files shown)


src/validation/validation.hpp
```diff
@@ -123,6 +125,11 @@ void create_synthetic_rsf_tab(View5D &rsf_tab, const int nw, const int nump,
                               const int numsza, const int numcolo3,
                               const int numalb, Real *synthetic_values);
 
+void convert_1d_std_to_2d_view_device(const std::vector<Real> &pmid_db,
```
jeff-cohere, line 128: Can you put comments in front of these functions that indicate they're copying to and from vectors? I know it's pretty obvious, but the `const` in front of `var_device` in this function might confuse some people. Also, consider replacing `std` with `vector` (though I think it's fine to keep it as `std` if that's what you decide).

jeff-cohere (approved):
Looks great. I left a few minor suggestions.

src/validation/validation.hpp
odiazib: Done.

singhbalwinder (approved):
Looks good! Thanks @odiazib !

tangshuaiqi (approved):
Looks good to me.

Issue #175, opened 2023-05-24
Wet Deposition - clddiag FIXME


CameronRutherford:
Issue could probably do with a better name, but in porting clddiag (see - https://github.com/eagles-project/mam4xx/pull/164), a FIXME noted that `evapc` should potentially be used in the `lprec` calculation.

See [wet_dep.hpp](https://github.com/eagles-project/mam4xx/tree/main/src/mam4xx/wet_dep.hpp) (link will work once #164 is merged).

jeff-cohere:
@overfelt FYI

PR #235, opened 2023-08-28
Add wetdep compute_tendencies test.

overfelt:


jeff-cohere (approved):
Looks good to me. Thanks!

Not-very-related question: Is #175 still an issue that we need to address?

overfelt:
Yes, I would say that #175  is still an issue. It is more of a question than an issue: "Should evapc be used in the lprec calculation?"  Until someone comes up with a definitive answer, I guess it stays open.

PR #237, opened 2023-08-29
Add a page to the docs explaining how to get `clang-format` v14 via Homebrew

mjs271:
Added a cleaned up version of the message I previously sent in slack. I retained the bit about `direnv` since it could be useful, but not sure if it belongs here and can ditch it if not. Also added bash syntax specifiers to a few code blocks that I noticed while scrolling around.


jeff-cohere (approved):
This looks great. Thanks Mike!

odiazib (approved):
Thanks, Mike!

PR #244, opened 2023-09-07
Adding a Sources container for storing surface and volume sources.

jeff-cohere:
For now, I'm just adding the container. When it's appropriate, we can add sources to the appropriate interfaces.  On the mam4xx side, we can assume that e.g. emissions data are placed properly within Sources. On the EAMxx side, we must make sure that such source data are properly interpolated from files, etc.


odiazib (approved):

jeff-cohere:
It worries me a little bit that we might have to carry around over 30 views just for source terms, but this should get us started.

singhbalwinder (approved):
Looks good!

PR #246, opened 2023-09-13
Fix indexing of qaer array in gasaerexch_soaexch.hpp.

overfelt:
The qaer array should be indexed with aerosol ids and not gas ids as are most of the arrays in these functions. There is an array, gas_to_aer, that maps between the two.

This was not needed before because for the MAM4 model we are using the AeroId SOA and GasId SOAG are both 0 so it did not matter that that gas was used to index the aerosol.


jeff-cohere (approved):
Awesome! Thanks for this, James.

PR #251, opened 2023-09-19
het_diags from mo_chm_diags

jaelynlitz:
isolating this completed function from still in progress chm_diags


src/validation/mo_chm_diags/het_diags.cpp
```diff
@@ -0,0 +1,109 @@
...
+    auto wrk_wd_host = View1DHost(vector0.data(), gas_pcnst);
+    const auto wrk_wd = View1D("wrk_wd", gas_pcnst);
+    Kokkos::deep_copy(wrk_wd, wrk_wd_host);
+
+    auto sox_wk_host = View1DHost(single_vector0.data(), 1);
+    const auto sox_wk = View1D("sox_wk", 1);
+    Kokkos::deep_copy(sox_wk, sox_wk_host);
+
+    const Real sox_species[3] = {4, -1, 3};
+    const Real adv_mass[gas_pcnst] = {
```
jeff-cohere, line 80: Is this not included in the Skywalker test data? Should it be? Asking out of curiosity.

jeff-cohere (approved):
Looks good! I asked one question.

src/validation/mo_chm_diags/het_diags.cpp
jaelynlitz: It was not in the validation data - I pulled it from [here](https://github.com/eagles-project/e3sm_mam4_refactor/blob/442cbca05f8f85aef326b46a53b2c5f0efd207d0/components/eam/src/chemistry/pp_linoz_mam4_resus_mom_soag/mo_sim_dat.F90#L42)

jeff-cohere: Aha, that makes sense. It's generated and specific to the chemical mechanism. Thanks!

mjs271 (approved):
Looks good to me!

PR #254, opened 2023-09-27
Fixes for building mam4xx within EAMxx on GPUs

jeff-cohere:


odiazib (approved):

overfelt:
The default constructor for the Config() sub-class of nucleation was made deviceable, maybe the same for the destructor and copy constructors.

jeff-cohere:
> The default constructor for the Config() sub-class of nucleation was made deviceable, maybe the same for the destructor and copy constructors.

That's a good point. Let me patch that in.

overfelt (approved):

PR #255, opened 2023-09-28
More GPU-related fixes

jeff-cohere:
This PR uses the new GPU-sensible `min()`, `max()`, and `epsilon()` functions offered by Haero. It also annotates some more functions with `KOKKOS_INLINE_FUNCTION` to make them callable on-device.

### DIFF
--- a/src/mam4xx/nucleation.hpp
+++ b/src/mam4xx/nucleation.hpp
@@ -649,6 +649,7 @@ public:
 
   // init -- initializes the implementation with MAM4's configuration and with
   // a process-specific configuration.
+  KOKKOS_INLINE_FUNCTION
   void init(const AeroConfig &aero_config,
             const Config &nucl_config = Config()) {
     // Set nucleation-specific config parameters.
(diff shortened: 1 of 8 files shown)


src/mam4xx/nucleation.hpp
```diff
@@ -649,6 +649,7 @@ class Nucleation {
 
   // init -- initializes the implementation with MAM4's configuration and with
   // a process-specific configuration.
+  KOKKOS_INLINE_FUNCTION
   void init(const AeroConfig &aero_config,
```
overfelt, line 653: In the Nucleation class it is possible for init() to be device callable, but in the ConvProc class I do a lot of resizing of Kokkos Views inside of the init() function since that is where the number of levels in the atmosphere gets set.  In general is init() supposed to be device callable?

overfelt (approved):

src/mam4xx/nucleation.hpp
jeff-cohere: Great question. I'm not sure that `init` has to be device-callable in the long run. At the moment, the need for this is an artifact of the structure of the directly-ported `modal_aero_amicphys` subroutine. So this change might be a temporary fix needed to get things running before we restructure things to be simpler.

mjs271 (approved):

PR #261, opened 2023-11-16
Adding some basic data types to simplify photolysis rate calculations

jeff-cohere:
This PR adds a column-independent struct called `PhotoTableData` that stores all the data used to interpolate photolysis rates, and a column-specific struct called `PhotoTableWorkArrays` that contains views that function as work arrays. These two types encapsulate array data that once lived in Fortran module variables. I find it too hard to work with functions with 87 arguments, so this is how I've been dealing with them. :-)

No logic is changed, apart from the order of allocation of views. I hope this will allow me to simplify the photolysis code in our EAMxx aerosol microphysics process.


odiazib (approved):
It looks great! Thanks for cleaning this code.

PR #262, opened 2023-11-20
Changed phtcnt from a Real to an int.

jeff-cohere:
This change lets us use `phtcnt` to size arrays on the GPU.


odiazib (approved):
Thanks for fixing this bug.

PR #265, opened 2023-11-27
Added a few more constants to `mam4::gas_chemistry` namespace

jeff-cohere:
Also, we need to use `ConstColumnView` when we refer to column views that are taken from the `Atmosphere` type. This is an unfortunate example of how C++ pushes really hard on the `const` idea and falls short. It shouldn't be necessary to create a whole new type in the name of constness, but this wart has been around since the advent of iterators in the STL.

</rant>

### DIFF
--- a/.github/workflows/auto_test.yml
+++ b/.github/workflows/auto_test.yml
@@ -22,9 +22,6 @@ jobs:
 
     runs-on: ${{ matrix.os }}
 
-#    # Pre-fab container with third-party libs installed.
-#    container: coherellc/haero-tpl:${{ matrix.build-type }}-${{ matrix.fp-precision }}
-
     # Environment variables
     env:
       CI: 1   # indicates that we are running in a CI environment.


jeff-cohere:
🤔 GitHub seems to be struggling to determine whether we can automatically merge this branch (Dear GitHub: we can!). Hopefully it clears up sometime soon.

jeff-cohere:
Well, for what it's worth, I've built and tested these changes in double precision on CPU and GPU configurations for my local Linux workstation. If this issue persists, I'll just manually merge and we can hope that GitHub gets its wires uncrossed in the near future.

mjs271 (approved):

jeff-cohere:
Hrmm, it's not a browser issue (I'm writing this comment on another browser that I haven't used for anything recently). Also, my most recent commit is missing. Maybe they're putting "AI" in charge of these things now?

PR #267, opened 2023-11-29
Ported set_ub_col and setcol

jeff-cohere:
Not many chemistry subroutines left after this!

### DIFF
--- a/src/mam4xx/mo_photo.hpp
+++ b/src/mam4xx/mo_photo.hpp
@@ -3,6 +3,7 @@
 
 #include <haero/math.hpp>
 #include <mam4xx/aero_config.hpp>
+#include <mam4xx/gas_chem_mechanism.hpp>
 #include <mam4xx/mam4_types.hpp>
 #include <mam4xx/utils.hpp>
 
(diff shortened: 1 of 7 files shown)


jeff-cohere:
The corresponding [mam_x_validation PR](https://github.com/eagles-project/mam_x_validation/pull/54) should be reviewed before this one.

jeff-cohere:
Okay, I see a GPU failure that needs to be addressed...

mjs271 (approved):
Looks good to me. Pushed a small typo correction, only because I sat scratching my head for a minute before I figured it out 🙃

src/mam4xx/mo_photo.hpp
```diff
@@ -259,6 +260,41 @@ void cloud_mod(const Real zen_angle, const Real *clouds, const Real *lwc,
...
+
+  // NOTE: o3_exo_col is 0 because O2 does not appear in the solsym array.
+  // NOTE: Further, it's only used to initialize the 0th vertical level, so
+  // NOTE: we ignore it in this single-level calculation.
+
+  // the following logic was extracted from the calc_col_delta subroutine under
+  // the above conditions that o3_ndx == 0, o3_inv_ndx == -1, and O2 is not
+  // a species or an invariant of interest
+  constexpr Real xfactor = 2.8704e21 / (9.80616 * 1.38044); // BAD_CONSTANT!
+  constexpr int o3_ndx = 0;
```
odiazib, line 282: I would make this index an input. In case, someone changes the order of the species.

odiazib (approved):
It looks great! just one minor comment.

src/mam4xx/mo_photo.hpp
jeff-cohere: This species ordering is encoded in the preprocessor-generated chemical mechanism, so I'm a little reluctant to imply that it's an adjustable parameter. We'll have to make a lot of code changes to support a more general chemical mechanism as it is. What do you think?

odiazib: I would make only one small change:

```
KOKKOS_INLINE_FUNCTION
void set_ub_col(Real &o3_col_delta,
                const int o3_ndx, // add this index as input.
                const Real vmr[mam4::gas_chemistry::gas_pcnst],
                const Real invariants[mam4::gas_chemistry::nfs],
                const Real pdel)

```

jeff-cohere: I've thought about this a bit more, and I still don't think we should pretend that `o3_ndx` is able to be changed. There's another index  (`o3_inv_ndx`) that can be set if O3 happens to be an invariant. Both of these parameters are tied to the chemical mechanism, which is etched into mam4xx. To me, the comments are enough for anyone to realize the limitations of our current approach. And I think we need to rewrite this and other things as part of any enhancement to the chemical mechanism.

odiazib: We need to defined parameters like o3_indx in one file; for example, `gas_chem_mechanism.hpp`.  But, I will add this to a wish list.

jeff-cohere: That makes sense. But even if we do that, we can refer directly to the index as a constant instead of putting it as a parameter.

Thanks for the discussion! These things are important.

PR #268, opened 2023-12-05
Add `get_dtdz` from tropopause

jaelynlitz:


jaelynlitz:
separated `get_dtdz` from #258. Merged test files are passing with tolerance of 9e-3

src/mam4xx/tropopause.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+using ConstColumnView = haero::ConstColumnView;
+// From radconstants
+constexpr int nswbands = 14;
+constexpr int nlwbands = 16;
+using View2D = DeviceType::view_2d<Real>;
+constexpr Real km_inv_to_m_inv = 0.001; // 1/km to 1/m
+
+constexpr Real shr_const_rgas =
+    haero::Constants::r_gas * 1e3; // Universal gas constant ~ J/K/kmole
```
jeff-cohere, line 22: We've been using the notation `[J/K/kmole]` to indicate the units for a constant or quantity.

src/validation/tropopause/get_dtdz.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+void get_dtdz(Ensemble *ensemble) {
+  ensemble->process([=](const Input &input, Output &output) {
+    const Real pm = input.get_array("pm")[0];
+    const Real pmk = input.get_array("pmk")[0];
+    const Real pmid1d_up = input.get_array("pmid1d_up")[0];
+    const Real pmid1d_down = input.get_array("pmid1d_down")[0];
+    const Real temp1d_up = input.get_array("temp1d_up")[0];
+    const Real temp1d_down = input.get_array("temp1d_down")[0];
+    // const Real cnst_kap = input.get_array("cnst_kap")[0];
```
jeff-cohere, line 23: Can we delete these two lines, or do you think we might need to re-add them later?

jaelynlitz: `cnst_kap`, `cnst_faktor`, and `cnst_ka1` are currently constants at the top of `tropopause.hpp`, but are also provided in the input file. I have opted to calculate them as constants in the hpp file but could change that

src/mam4xx/tropopause.hpp
```diff
@@ -0,0 +1,64 @@
...
+    haero::Constants::cp_dry_air; // specific heat of dry air   ~ J/kg/K
+
+constexpr Real cnst_kap =
+    (shr_const_rgas / shr_const_mwdair) / shr_const_cpdair; //  ! R/Cp
+
+constexpr Real cnst_faktor =
+    -haero::Constants::gravity /
+    haero::Constants::r_gas_dry_air; // acceleration of gravity ~ m/s^2/Dry air
+                                     // gas constant     ~ J/K/kg
+constexpr Real cnst_ka1 = cnst_kap - 1.0;
```
jeff-cohere, line 35: Are the above constants going to be used outside of the `get_dtdz` function? If any of them are specific to that function, they can be placed inside it.

src/mam4xx/tropopause.hpp
```diff
@@ -0,0 +1,64 @@
...
+#include <fstream>
+#include <iostream>
+
+namespace mam4 {
+
+namespace tropopause {
+
+using ConstColumnView = haero::ConstColumnView;
+// From radconstants
+constexpr int nswbands = 14;
```
odiazib, line 16: Add BAD CONSTANT

src/validation/tropopause/get_dtdz.cpp
jeff-cohere: I think your intuition is good here. Let's keep constants, uhh, constant, and not expose them as variable parameters. :-) Okay to delete these lines in my opinion.

jeff-cohere (approved):
Looks good! Thanks. Left a few minor comments.

odiazib (approved):
It looks great!

src/mam4xx/tropopause.hpp
jaelynlitz: They will also be in the function `twmo` in tropopause

jaelynlitz:
Made the suggested changes - merging! 😄

PR #272, opened 2023-12-19
Fixes for Scream integration.

overfelt:
Scream does strong const checking of Kokkos views so modified the input only views of dropmixnuc to be const data.

Fix an index out-of-bounds error.

### DIFF
--- a/src/mam4xx/ndrop.hpp
+++ b/src/mam4xx/ndrop.hpp
@@ -792,29 +792,33 @@ void update_from_cldn_profile(
     Real &qcld, Real factnum_col[AeroConfig::num_modes()],
     Real &eddy_diff, // out
     Real nact[AeroConfig::num_modes()], Real mact[AeroConfig::num_modes()]) {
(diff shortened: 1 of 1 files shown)


odiazib (approved):
It looks good to me.

odiazib:
I do not see the test that runs in i/gitlab/code.pnnl.gov. @jeff-cohere

jeff-cohere:
I think this was merged before those tests could report, so I don't know if we can do anything about it on this end. You can probably log into the [PNNL CI system](http://gitlab.pnnl.gov/), though, if you want to check the results there.

overfelt:
Oh, thought I saw a green check mark that the tests had finished.

jeff-cohere:
Okay, it looks like the pipeline is still running--we're just not getting the results reported. Here's the link to the pipeline (requires PNNL single-signon): https://code.pnnl.gov/e3sm/eagles/mam4xx/-/pipelines

PR #273, opened 2023-12-22
`mo_sethet` pt 1 - port of `sethet` dependencies 

jaelynlitz:
Port from [mo_sethet](https://github.com/eagles-project/e3sm_mam4_refactor/blob/refactor-maint-2.0/components/eam/src/chemistry/mozart/mo_sethet.F90) - `sethet`'s dependencies

Separating the dependencies from `sethet` bc this PR is already big 😄 

- [X] `calc_het_rates`
- [X] `calc_precip_rescale`
- [X] `find_ktop`
- [X] `gas_washout`


src/mam4xx/mo_sethet.hpp
```diff
@@ -0,0 +1,177 @@
...
+  Real const0 = boltz_cgs * 1.0e-6; // [atmospheres/deg k/cm^3]
+  Real geo_fac = 6.0; // geometry factor (surf area/volume = geo_fac/diameter)
+  Real xrm = .189;    // mean diameter of rain drop [cm]
+  Real xum = 748.0;   // mean rain drop terminal velocity [cm/s]
+
+  //-----------------------------------------------------------------
+  //       ... calculate the saturation concentration eqca
+  //-----------------------------------------------------------------
+  for (int kk = 0; kk < plev;
+       kk++) { // FIXME: not sure if this should be a Kokkos for or not...
```
jaelynlitz, line 123: @odiazib @jeff-cohere @mjs271 looking for input here if this should be a Kokkos for or not?

jeff-cohere: The rule we've been using is this: if all the levels (data related to the index k) are independent, then it's safe to use a Kokkos parallel_for.

jaelynlitz: Okay I put the first part in a parallel_for then the second part in a regular for because it does a check after adding each element to a sum - not sure if that can be done in a parallel_reduce... but closer regardless!

src/mam4xx/mo_sethet.hpp
```diff
@@ -0,0 +1,181 @@
...
+    if (allca < xeqca(kk)) {
+      xgas(kk) = haero::max(xgas(kk) - xca(kk), 0.0);
+    }
+  }
+} // end subroutine gas_washout
+
+//=================================================================================
+KOKKOS_INLINE_FUNCTION
+void find_ktop(
+    Real rlat,        // latitude in radians for columns
```
jaelynlitz, line 155: Looking ahead to `sethet` - this rlat value comes from calling [`get_rlat_all_p`](https://github.com/eagles-project/e3sm_mam4_refactor/blob/f6f291b80b9e245a04d422e7bcda5a9609c12107/components/eam/src/physics/cam/phys_grid.F90#L2140). For the C++, should I instead expect `rlat` to come as a parameter from EAMxx?

jeff-cohere: Yes, I think that's reasonable.

src/mam4xx/mo_sethet.hpp
```diff
@@ -0,0 +1,181 @@
...
+                    Real &het_rates) // rainout loss rates [1/s]// out
+{
+  //-----------------------------------------------------------------
+  // calculate het_rates
+  // input arguments are different for different species
+  //-----------------------------------------------------------------
+  Real work3;
+  Real h2o_mol = 1.0e3 / mass_h2o; // [gm/mol water]
+  // FIXME: BAD CONSTANT should haero::Constants::molec_weight_h2o be used
+  // somewhere here instead?
```
jeff-cohere, line 50: Perhaps (and maybe even "probably"), but marking this as a `BAD CONSTANT` is sufficient for now.

src/validation/mo_sethet/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+    gas_washout_ts_355
+    )
+# # matching the tests and errors, just for convenience
+
+set(DEFAULT_TOL 1e-13)
+set(ERROR_THRESHOLDS
+    ${DEFAULT_TOL}
+    ${DEFAULT_TOL}
+    ${DEFAULT_TOL} # output is an int
+    9e-3
```
jeff-cohere, line 45: This tolerance is quite a bit higher than the default. Is there something fancy going on in the `gas_washout` logic? Are you confident that we're not missing something there?

jeff-cohere (approved):
This looks good to me. Have you tested this on a GPU? We're not getting reports back from Deception in our GitHub Actions, so we've been having to check GPU results ourselves. (And btw, if you can help with the GPU test reporting in that regard, we'd be most grateful!)

I left one comment about the relatively high tolerance needed for the gas washout validation test.

src/mam4xx/mo_sethet.hpp
```diff
@@ -0,0 +1,181 @@
...
+#include <mam4xx/mam4_types.hpp>
+#include <mam4xx/mo_chm_diags.hpp>
+#include <mam4xx/utils.hpp>
+
+namespace mam4 {
+
+namespace mo_sethet {
+
+// const int ktop = ConvProc::Config::ktop;
+// int ktop = 47; // BAD_CONSTANT: only true for nlev == 72
```
odiazib, line 15: Do we need this line?

src/mam4xx/mo_sethet.hpp
```diff
@@ -0,0 +1,181 @@
...
+const Real pi = haero::Constants::pi;
+constexpr Real rgrav =
+    mo_chm_diags::rgrav; // reciprocal of acceleration of gravity ~ m/s^2
+constexpr const int gas_pcnst = gas_chemistry::gas_pcnst;
+const Real boltz_cgs = haero::Constants::boltzmann * 1.e7; // erg/K
+// number of vertical levels
+constexpr const int pver = mam4::nlev;
+
+// FIXME: BAD CONSTANT
+constexpr const Real mass_h2o = 18.0;   // mass of water vapor [amu]
```
odiazib, line 26: Is this molecular weight of H2O? We could use haero::Constant.

src/mam4xx/mo_sethet.hpp
```diff
@@ -0,0 +1,181 @@
...
+      satf * haero::max(rain / (h2o_mol * (work1 + 1.0 / (xhen * work2))), 0.0);
+  het_rates = work3 + tmp_hetrates;
+
+} // end calc_het_rates
+//=================================================================================
+
+//=================================================================================
+KOKKOS_INLINE_FUNCTION
+void calc_precip_rescale(
+    ColumnView cmfdqr,   // dq/dt for convection [kg/kg/s] //in
```
odiazib, line 62: I believe, we need to use `const ColumnView&` instead of `ColumnView& `.

src/mam4xx/mo_sethet.hpp
```diff
@@ -0,0 +1,181 @@
...
+
+} // end subroutine calc_precip_rescale
+
+//=================================================================================
+KOKKOS_INLINE_FUNCTION
+void gas_washout(const ThreadTeam &team,
+                 int plev,          // calculate from this level below //in
+                 Real xkgm,         // mass flux on rain drop //in
+                 Real xliq_ik,      // liquid rain water content [gm/m^3] // in
+                 ColumnView xhen_i, // henry's law constant
```
odiazib, line 104: I believe, we need to use `const ColumnView&` instead of `ColumnView& `.

src/mam4xx/mo_sethet.hpp
```diff
@@ -0,0 +1,181 @@
...
+      precip(kk) = precip(kk) * total_rain / total_pos;
+    }
+  }
+
+} // end subroutine calc_precip_rescale
+
+//=================================================================================
+KOKKOS_INLINE_FUNCTION
+void gas_washout(const ThreadTeam &team,
+                 int plev,          // calculate from this level below //in
```
odiazib, line 101: I would add `const`

odiazib (changes requested):

src/validation/mo_sethet/CMakeLists.txt
jaelynlitz: The answers get increasingly larger - up to around 1e10 - the answers look visually close to the extent that they are printed so I think the logic is in line with the fortran

jaelynlitz:
> This looks good to me. Have you tested this on a GPU? We're not getting reports back from Deception in our GitHub Actions, so we've been having to check GPU results ourselves. (And btw, if you can help with the GPU test reporting in that regard, we'd be most grateful!)
> 
> I left one comment about the relatively high tolerance needed for the gas washout validation test.

I will test on GPU and look into the Deception CI issues next!

src/mam4xx/mo_sethet.hpp
jaelynlitz: looks like it could be the molecular weight of water in g/mol instead of kg/mol?

jaelynlitz:
I think this is ready pending pipelines and another review 😄

src/mam4xx/mo_sethet.hpp
```diff
@@ -0,0 +1,179 @@
...
+// number of vertical levels
+constexpr const int pver = mam4::nlev;
+
+// FIXME: BAD CONSTANT
+// mass of water vapor [amu] //convert to g/mol from kg/mol
+constexpr const Real mass_h2o = haero::Constants::molec_weight_h2o * 1000;
+constexpr const Real cm3_2_m3 = 1.0e-6; // convert cm^3 to m^3
+constexpr const Real liter_per_gram = 1.0e-3;
+constexpr const Real avo2 =
+    avo * liter_per_gram * cm3_2_m3; // [liter/gm/mol*(m/cm)^3]
```
mjs271, line 30: I've never seen this before, but is doubling up `constexpr const` a usual thing?

mjs271 (approved):
Looks great--nice work! 🚀

src/mam4xx/mo_sethet.hpp
jaelynlitz: I'll be honest, I didn't notice that I did that - I think `constexpr` is sufficient but apparently the extra `const` didn't make anything complain

jeff-cohere: That is a lotta `const`! Maybe it's good to remove the extra `const` just to prevent future `const`-fusion

src/mam4xx/mo_sethet.hpp
odiazib: it is g/mol. Thus, we will need to multiple by 1000 the values from  haero::Constant::molec_weight_h2o.  I also noted the values in haero is 
```
/// Molecular weight of water [kg/mol]
static constexpr Real molec_weight_h2o = 0.018016;
```
, which has a few extra digits.

jeff-cohere:
Just a heads up that we want to merge #282 before this PR because of an update to the Haero interface.

odiazib (approved):
It looks great!

jaelynlitz:
I'm not sure why the double precision builds are failing in the Deception pipelines, it compiles just fine when I do it manually

jaelynlitz:
Hopefully that's the last rebase and this can merge after CI passes 😄

jeff-cohere:
Merge away!

PR #275, opened 2024-01-03
calcsize - fix GPU segfault

odiazib:


jeff-cohere (approved):
I've confirmed that this fixes the error I observed. Thanks, Oscar!

PR #276, opened 2024-01-08
Changed 4D and 5D view type definitions in mo_photo.hpp so we can use subviews in EAMxx.

jeff-cohere:


odiazib (approved):
Looks good to me. 

I wonder if Kokkos::subview work.

PR #279, opened 2024-01-10
Long-awaited Port of `mo_setinv`

mjs271:
This PR ports `mo_setinv`, which sets the "invariant species" within [`mo_gas_phase_chemdr.F90`](https://github.com/eagles-project/e3sm_mam4_refactor/blob/f6f291b80b9e245a04d422e7bcda5a9609c12107/components/eam/src/chemistry/mozart/mo_gas_phase_chemdr.F90#L375).

There are some idiosyncratic constants and indexing schemes that are used on the fortran side, and I'm not sure some of these will make it into EAMxx. For example, there's some convoluted combination of strings and indices used to get the `cnst_offline` quantities from some faraway source. So, I've stored these variables in a `Config` struct for now that's initialized in `setinv()` and passed as an argument to `setinv_single_level()`. 

There are two validation tests, `setinv_test_single_level.cpp` that runs data for a number of time steps, and `setinv_test_nlev.cpp` that duplicates a single one of these time-step inputs and runs that condition on a full column in a `parfor()` (the same workflow as used for `mo_setsox`).

### DIFF
--- a/src/mam4xx/mo_setinv.hpp
+++ b/src/mam4xx/mo_setinv.hpp
@@ -20,8 +20,7 @@ struct Config {
   // FIXME: BAD CONSTANTS!
   // conversion factor for Pascals to dyne/cm^2
   Real Pa_xfac = 10.0;
-  // presumably, the boltzman constant, in CGS units
-  // NOTE: passed in via yaml input file for now
+  // presumably, the boltzmann constant, in CGS units
   Real boltz_cgs = 0.13806500000000001E-015;
   // NOTE: the indices (xyz_ndx) are translated to zero-indexed in the
   // constructor
--- a/src/mam4xx/mo_setinv.hpp
+++ b/src/mam4xx/mo_setinv.hpp
@@ -1,0 +1,123 @@
+#ifndef MAM4XX_MO_SETINV_HPP
+#define MAM4XX_MO_SETINV_HPP
+
+#include <haero/atmosphere.hpp>
+
+#include <mam4xx/aero_config.hpp>
+#include <mam4xx/mam4_types.hpp>
+
+namespace mam4 {
+namespace mo_setinv {
+// number of invariants
+static const int nfs = 8;
+static const int num_tracer_cnst = 4;
+
+struct Config {
+  // FIXME: BAD CONSTANTS!
+  // conversion factor for Pascals to dyne/cm^2
+  Real Pa_xfac = 10.0;
+  // presumably, the boltzman constant, in CGS units
+  // NOTE: passed in via yaml input file for now
+  Real boltz_cgs = 0.13806500000000001E-015;
+  // NOTE: the indices (xyz_ndx) are translated to zero-indexed in the
+  // constructor
+  bool has_n2 = true;
+  int m_ndx = 0;
+  int n2_ndx = 1;
+  bool has_o2 = true;
+  int o2_ndx = 2;
+  bool has_h2o = true;
+  int h2o_ndx = 3;
+  // FIXME: this will likely need to be changed to match the order of how the
+  // "cnst_offline" species are passed in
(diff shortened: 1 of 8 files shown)


jeff-cohere:
Thanks, @mjs271 ! Have you tested this branch on a GPU?

src/mam4xx/mo_setinv.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+#include <haero/atmosphere.hpp>
+
+#include <mam4xx/aero_config.hpp>
+#include <mam4xx/mam4_types.hpp>
+
+namespace mam4 {
+namespace mo_setinv {
+// number of invariants
+static const int nfs = 8;
```
jeff-cohere, line 12: FYI, `nfs` is available as a `constexpr` in the `mam4::gas_chemistry` namespace. See `gas_chem_mechanism.hpp`.

mjs271:
> EDIT: I've verified that it passes all tests on GPU. Well done!

Thank you!

src/mam4xx/mo_setinv.hpp
mjs271: Ah, good catch! I gave that a shot previously, and it was crashing my builds for some reason, but that may have been related to something else--I'll swap that out!

src/mam4xx/mo_setinv.hpp
```diff
@@ -0,0 +1,123 @@
...
+static const int num_tracer_cnst = 4;
+
+struct Config {
+  // FIXME: BAD CONSTANTS!
+  // conversion factor for Pascals to dyne/cm^2
+  Real Pa_xfac = 10.0;
+  // presumably, the boltzman constant, in CGS units
+  // NOTE: passed in via yaml input file for now
+  Real boltz_cgs = 0.13806500000000001E-015;
```
jeff-cohere, line 22: I'm not wild about making this a configurable parameter, but that hardly distinguishes this particular piece of code!

jeff-cohere (approved):
This looks good to me, but I'm interested to hear what others think.

src/mam4xx/mo_setinv.hpp
```diff
@@ -0,0 +1,124 @@
...
+  constexpr int nk = mam4::nlev;
+
+  Kokkos::parallel_for(
+      Kokkos::TeamThreadRange(team, nk), KOKKOS_LAMBDA(int k) {
+        const Real tfld_k = tfld(k);
+        const Real h2ovmr_k = h2ovmr(k);
+        const Real pmid_k = pmid(k);
+        Real invariants_k[nfs];
+        const int nspec = AeroConfig::num_gas_phase_species();
+        Real vmr_k[nspec];
```
odiazib, line 107: I would use a view_2d for vrm and employee subview to get vrm_k.

src/mam4xx/mo_setinv.hpp
```diff
@@ -0,0 +1,124 @@
...
+        for (int i = 0; i < nspec; ++i) {
+          vmr_k[i] = vmr[i](k);
+        }
+        Real cnst_offline_k[num_tracer_cnst];
+        for (int i = 0; i < num_tracer_cnst; ++i) {
+          cnst_offline_k[i] = cnst_offline[i](k);
```
odiazib, line 113: I would also use a view_2d and subview to avoid copy data.

odiazib (approved):
It looks good to me.

src/mam4xx/mo_setinv.hpp
mjs271: Turns out the `NOTE` was an extraneous leftover--not passed in via yaml.

The local cgs Boltzman constant, on the other hand... 😶

jaelynlitz (approved):

PR #281, opened 2024-01-16
Fix GitLab Pipeline Status 

jaelynlitz:
Updating docs with the steps I took to get the GitLab status to reappear in GitHub

Original issue:
GitLab status stopped showing up in GitHub

In GitLab > Settings > Integrations > GitHub we were getting this error
```
Connection failed. Check your integration settings. POST https://api.github.com/repos/eagles-project/mam4xx/statuses/9d1e78201b48b1c6cdc09ca4a70e49cba63df58f: 401 - Bad credentials // See: https://docs.github.com/rest
```

I then generated a PAT with write and admin:repo_hook permissions, went back to that setting and entered my new token and that has gotten the pipeline status to reappear

This fix will last until next winter.


jaelynlitz:
I also gained `Owner` permissions on the GitLab to take this over for @cameronrutherford

jeff-cohere:
Hey @jaelynlitz . Thanks for putting this together. If I had to guess what's going wrong with the checks, I'd say we need to rebuild Haero on Deception and friends, either by submitting a commit message with `[rebuild-haero]` or manually.

jaelynlitz:
> Hey @jaelynlitz . Thanks for putting this together. If I had to guess what's going wrong with the checks, I'd say we need to rebuild Haero on Deception and friends, either by submitting a commit message with `[rebuild-haero]` or manually.

copy that, that's what I suspected ✌️

jaelynlitz:
> > Hey @jaelynlitz . Thanks for putting this together. If I had to guess what's going wrong with the checks, I'd say we need to rebuild Haero on Deception and friends, either by submitting a commit message with `[rebuild-haero]` or manually.
> 
> copy that, that's what I suspected ✌️

Ah, finally rebuilt haero locally and that rebuild is causing the compilation error...

jeff-cohere:
> > > Hey @jaelynlitz . Thanks for putting this together. If I had to guess what's going wrong with the checks, I'd say we need to rebuild Haero on Deception and friends, either by submitting a commit message with `[rebuild-haero]` or manually.
> > 
> > 
> > copy that, that's what I suspected ✌️
> 
> Ah, finally rebuilt haero locally and that rebuild is causing the compilation error...

Do you mean you're getting that compilation error in the rebuild of Haero? Have you pulled the latest version of Haero? I believe @overfelt recently modified the constructor that's causing the issue.

jaelynlitz:
> > > > Hey @jaelynlitz . Thanks for putting this together. If I had to guess what's going wrong with the checks, I'd say we need to rebuild Haero on Deception and friends, either by submitting a commit message with `[rebuild-haero]` or manually.
> > > 
> > > 
> > > copy that, that's what I suspected ✌️
> > 
> > 
> > Ah, finally rebuilt haero locally and that rebuild is causing the compilation error...
> 
> Do you mean you're getting that compilation error in the rebuild of Haero? Have you pulled the latest version of Haero? I believe @overfelt recently modified the constructor that's causing the issue.

Yes, I rebuilt haero ~10 mins ago and reproduced the error locally

overfelt:
Ah, yes this is my fault.   I'll submit a revert of that commit.

jeff-cohere:
Here's what I found in the build log for single-precision builds:

```
...
.../haero_Debug_single/src/mam4xx/modal_aer_opt.hpp(47): error: floating-point value does not fit in required floating-point type
...
```

jaelynlitz:
> Here's what I found in Deception's build log for single-precision builds:
> 
> ```
> ...
> .../haero_Debug_single/src/mam4xx/modal_aer_opt.hpp(47): error: floating-point value does not fit in required floating-point type
> ...
> ```

seems like 1e-60 doesn't fit in a single precision Real?

jeff-cohere:
Could be an oddity of "CUDA C++", because this C++ program compiles fine:

```
#include <iostream>

int main() {
  float v = 1.e-60;
  std::cout << "v = " << v << std::endl;
  return 0;
}
```

jaelynlitz:
> Could be an oddity of "CUDA C++", because this C++ program compiles fine:
> 
> ```
> #include <iostream>
> 
> int main() {
>   float v = 1.e-60;
>   std::cout << "v = " << v << std::endl;
>   return 0;
> }
> ```
> 
> Update: found some chatter on the Intarwebs about certain compilers not accepting values less than `FLT_MIN` (~1e-38) for single-precision `float`. Maybe `nvcc` is one of those compilers. In which case we must replace the offending constant with a more agreeable value.

Interesting - 1e-40 (small_value_40) seems to be compiling, could I just replace the small_value_60 with 40 if tests pass? It is used in one comparison in modal_aer_opt

jeff-cohere:
> Interesting - 1e-40 (small_value_40) seems to be compiling, could I just replace the small_value_60 with 40 if tests pass? It is used in one comparison in modal_aer_opt

I think this is a good solution. @overfelt, any objections?

overfelt:
I have no objection to using  1e-40.

jeff-cohere (approved):
Thanks for putting this together, Jaelyn! Please merge when ready.

overfelt (approved):

PR #291, opened 2024-02-06
Refactor the main loops within Wet Deposition.

overfelt:
Wet deposition has loops within loops within loops
with column levels within species within phase
within modes. Previously I had tried to put the outer loop
to be the levels as this was the approach at the time.
This resulted in awkward code and restrictions
on the number of threads that could be run.

Now the inner loop is over levels with outer loops are over
modes, phases and species.  This requires a lot of parallel_for
loops on teams but since it is all within compute_tendencies
which is already on device, these are not a kernel launches
and should be just as fast as before.

### DIFF
--- a/src/mam4xx/wet_dep.hpp
+++ b/src/mam4xx/wet_dep.hpp
@@ -913,6 +913,9 @@ void wetdepa_v2(const Real deltat, const Real pdel, const Real cmfdqr,
   const Real small_value_12 = 1.e-12;
   const Real small_value_36 = 1.e-36;
 
+  fracis = scavt = iscavt = icscavt = isscavt = bcscavt = bsscavt = rcscavt =
+      rsscavt = 0;
+
   // omsm = (1 - small number) used to prevent roundoff errors below zero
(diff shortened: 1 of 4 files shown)


src/mam4xx/wet_dep.hpp
```diff
@@ -1157,12 +1160,336 @@ void clddiag(const int nlev, const Real *temperature, const Real *pmid,
...
+      Kokkos::TeamThreadRange(team, nlev),
+      KOKKOS_LAMBDA(int k) { ptend_q(k, mm) += scavt[k]; });
+}
 } // namespace wetdep
 
 /// @class WedDeposition
 /// Wet Deposition process for MAM4 aerosol model.
 class WetDeposition {
 public:
+  static constexpr int gas_pcnst = 40;
```
singhbalwinder, line 1492: We should probably declare `40` in aero config so that all processes can share this number.

singhbalwinder (approved):
Looks great! The code looks very clean and nicely refactored!

PR #292, opened 2024-02-06
Add check to throw on divide by zero.

overfelt:
Prints a message about the Geopotential height not being decreasing as the ndrop algorithm divides by the difference in two adjacent levels.


singhbalwinder (approved):
Looks good. Thanks!

PR #293, opened 2024-02-07
Define gas_pcnst in aero_model and use that one everywhere else.

overfelt:


singhbalwinder (approved):
Thanks for fixing this!

PR #294, opened 2024-02-08
Allows mam4xx to be built as part of a larger CMake project

### LINKED ISSUE #290: Make mam4xx buildable within a subdirectory of a host project
(See [Haero issue 477](https://github.com/eagles-project/haero/issues/477) for more background)

In order to build a CIME-enabled E3SM case for testing our C++ port of MAM, we need to be able to incorporate mam4xx into EAMxx's build process in a way that freely passes build configuration parameters from EKAT -> Haero -> mam4xx. Currently, mam4xx assumes that haero is installed as a third-party library. This is fine, but we *also* need the ability for mam4xx to check for a haero library target within a unified build system.

This issue tracks the addition of this capability.

jeff-cohere:
This is the last missing piece of the puzzle for CIME-enabled builds. This PR should go in after the corresponding Haero PR, after which we can reset the haero and mam4xx submodules in our SCREAM fork to their respective `main` branches.

Closes #290

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,18 +3,19 @@ cmake_minimum_required(VERSION 3.17.0)
 message(STATUS "Configuring with build type: ${CMAKE_BUILD_TYPE}")
 
 # We rely on Haero for basic data structures and third-party libraries.
-if (NOT MAM4XX_HAERO_DIR)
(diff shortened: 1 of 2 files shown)


singhbalwinder (approved):
This looks good as well. Thanks!

PR #296, opened 2024-02-09
Oscar/aerosol optical depth

odiazib:


singhbalwinder (approved):
Looks good! I tried it with EAMxx global simulation and it works for me. These additions require changes to the optics interface code. Thanks for working on it on a short notice!

PR #297, opened 2024-02-15
Trivial change to use aero_model::gas_pcnst.

overfelt:
The classes that set a gas_pcnst from aero_model::gas_pcnst in public header data were changed to use the one from aero_model directly. Clean up the code a bit.

Note that there is a gas_pcnst in gas_chemistry that is set to 31. The one in aero_model is set to 40 and it might be a bit confusing.

### DIFF
--- a/src/mam4xx/aero_model.hpp
+++ b/src/mam4xx/aero_model.hpp
@@ -716,10 +716,10 @@ int lptr_nacl_a_amode(const int imode) {
   return lptr_nacl_a_amode[imode];
 }
 
-static constexpr int gas_pcnst = 40;
+static constexpr int pcnst = 40;
 KOKKOS_INLINE_FUNCTION
 int mmtoo_prevap_resusp(const int i) {
-  const int mmtoo_prevap_resusp[gas_pcnst] = {
+  const int mmtoo_prevap_resusp[pcnst] = {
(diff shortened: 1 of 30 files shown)


singhbalwinder:
You are right, it is very confusing to call this variable "gas_pcnst" when it has space to store gasses and other constituents. In E3SM or EAM, it is called just `pcnst`. Would it be too much work to change it from `gas_pcnt` to `pcnst`?

I reviewed the PR and it looks good to me.

overfelt:
It is easy to change aero_model::gas_pcnst to aero_model::pcnst.  Looking through the code we have
```
./modal_aer_opt.hpp:constexpr int pcnst = 40;
./water_uptake.hpp:constexpr int nvars = 40;
./ndrop.hpp:constexpr int nvars = 40;
./ndrop.hpp:constexpr int nvar_ptend_q = 40;

and the one we just created:
./aero_model.hpp:static constexpr int gas_pcnst = 40;
```
I assume these are different names for gss_pcnst and should also be replaced?

singhbalwinder:
Thanks, James! Yes, it would be great if we can unify them all into one and call it `pcnst`.

overfelt:
@singhbalwinder , seems the new changes to this pull request triggered a new review request when you have time.

singhbalwinder (approved):
Looks great! Thanks, James for taking care of this. This has been bothering me and I am happy that it is all consistent now.

PR #299, opened 2024-03-01
Oscar/water uptake

odiazib:
This PR fixed the EMAxx coupled test ``homme_shoc_mam_optics_cld_p3_rrtmgp``; see [EMAxx PR # 2718]( https://github.com/E3SM-Project/scream/pull/2718).   

This coupled test was failing in GPUs with the following error:

```
block: [70,0,0], thread: [0,69,0] Assertion `Kokkos contract violation:
    Expected precondition `r >= 0` evaluated false.
Error at scream/externals/ekat/extern/kokkos/core/src/Kokkos_Complex.hpp":692 
` failed.
```

Source of the error: if a real negative value is passed to  `Kokkos::sqrt`, this function will return a Nan. To fix this error, we must pass a complex value instead of real value in the `kokkos:sqrt` function.


jeff-cohere (approved):
Looks great. Thanks, Oscar. @pressel , this is an example of a complication that would have been avoided by using the Pete's bisection method.

PR #300, opened 2024-03-09
Fix calculate_gas_drydep_vlc_and_flux() test

mjs271:
This fixes the `calculate_gas_drydep_vlc_and_flux()` test, which is now passing with a tolerance of 1e-10.

It turns out that the input yaml files for the 4 sub-cases (that compose the single, merged test case) contained a small error. The variable `lcl_frc_landuse` in the input yaml file was for a 4-column chunk (4 columns x `n_land_type = 11`), rather than a single column like the remainder of the data. As a result, the test code was not reading the `lcl_frc_landuse` values that corresponded to the rest of the provided variables, since it was likely reading the first 11 values in the flattened array.

Additionally, the resulting `mam_calculate_gas_drydep_vlc_and_flux.py` validation file also appeared to be in error for the same reason (I hope 🤞). I found both of these bugs by running the standalone driver, and now the standalone driver and mam4xx produce matching output. I trust this, but I will also request new validation data from e3sm_mamrefactor, so we can be 100% sure.

### DIFF
--- a/src/mam4xx/mo_drydep.hpp
+++ b/src/mam4xx/mo_drydep.hpp
@@ -322,10 +322,8 @@ void calculate_resistance_rlux(
 
   // NOTE: as it stands, since rlux gets passed in, and we can't guarantee all
   // entries are initialized to 0, we do it here
-  for (size_t ispec = 0; ispec  < gas_pcnst; ++ispec)
-  {
-    for (size_t lt = 0; lt < n_land_type; ++lt)
-    {
+  for (size_t ispec = 0; ispec < gas_pcnst; ++ispec) {
+    for (size_t lt = 0; lt < n_land_type; ++lt) {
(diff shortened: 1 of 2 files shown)


overfelt (approved):

jeff-cohere (approved):
Excellent. Thanks, @mjs271 ! Looks like you just need to reformat.

jaelynlitz (approved):

jaelynlitz:
Looks like the PNNL pipelines are failing due to a haero mismatch - needs either a rebase and/or a rebuild of haero 😄 Thanks @mjs271 !

jeff-cohere:
Mike, you can amend your last commit and add `[rebuild-haero]` to the commit message on the first line to have the PNNL CI rebuild Haero.

PR #302, opened 2024-03-13
Enable FPEs for Testing

mjs271:
This is a "draft" PR that enables FPEs for testing and exposes some failures that were hiding.

Also, just to have it recorded somewhere, the FPE handling comes from EKAT, and the default cmake flags are

```
EKAT_ENABLE_FPE=ON
EKAT_ENABLE_FPE_DEFAULT_MASK=OFF
```

for which `DEFAULT_MASK=FE_INVALID | FE_INVALID | FE_OVERFLOW`.

So, I've added something to the tune of the following to the testing infrastructure

```
#ifndef NDEBUG
 ekat::enable_fpes(FE_INVALID | FE_INVALID | FE_OVERFLOW);
#endif
```

Note that I'm not marking it a draft so that the autotester will run tests and (hopefully) demonstrate the failures I'm seeing, namely:

#### Unit Tests
```
2 - mam4_nucleation_unit_tests (Failed)
3 - mam4_gasaerexch_unit_tests (Failed)
4 - mam4_coagulation_unit_tests (Failed)
11 - mam4_nucleate_ice_unit_tests (Failed)
20 - mode_averages (Failed)
```

#### Validation Tests
```
60 - run_stand_modal_aero_calcsize_sub (NUMERICAL)
396 - run_ccncalc_ts_1400 (NUMERICAL)
412 - run_stand_dropmixnuc_ts_1407 (NUMERICAL)
414 - run_dropmixnuc_ts_1400 (NUMERICAL)
416 - run_dropmixnuc_ts_1401 (NUMERICAL)
418 - run_dropmixnuc_ts_1402 (NUMERICAL)
420 - run_dropmixnuc_ts_1403 (NUMERICAL)
422 - run_dropmixnuc_ts_1404 (NUMERICAL)
424 - run_dropmixnuc_ts_1405 (NUMERICAL)
426 - run_dropmixnuc_ts_1406 (NUMERICAL)
428 - run_dropmixnuc_ts_1407 (NUMERICAL)
430 - run_dropmixnuc_ts_1408 (NUMERICAL)
432 - run_dropmixnuc_ts_1409 (NUMERICAL)
434 - run_dropmixnuc_ts_1410 (NUMERICAL)
436 - run_dropmixnuc_ts_1411 (NUMERICAL)
438 - run_dropmixnuc_ts_1412 (NUMERICAL)
440 - run_dropmixnuc_ts_1413 (NUMERICAL)
442 - run_dropmixnuc_ts_1414 (NUMERICAL)
444 - run_dropmixnuc_ts_1415 (NUMERICAL)
446 - run_dropmixnuc_ts_1416 (NUMERICAL)
448 - run_dropmixnuc_ts_1417 (NUMERICAL)  
```


overfelt:
So setting EKAT_ENABLE_FPE=ON in CMake sets the NDEBUG flag?   Ekat does not have its own flag for enabling FPEs?

jeff-cohere:
I think we're just assuming that we want FPEs for debug builds but not for optimized ones. To me, this makes sense, but we might want to allow a more flexible option.

mjs271:
> So setting EKAT_ENABLE_FPE=ON in CMake sets the NDEBUG flag? Ekat does not have its own flag for enabling FPEs?

Good question. As I've come to understand it, `EKAT_ENABLE_FPE=ON`, enables the **_possibility_** of catching FPEs, but doesn't add any of the possible types to the FPE mask that results in catching (trapping?) them during runtime.

Because ekat comes from haero, it seemed easiest to leave the default of `EKAT_ENABLE_FPE=ON` in haero and then, depending on the debug/release compile flag for mam4xx, that will determine whether we set the FPE masks within testing.

I'm certainly not married to this particular workflow, so let me know if there are preferred options.

overfelt:
I think FPEs can be set at runtime and will not effect performance if not turned on.  It seems that both EKAT_ENABLE_FPE=ON and a debug compile are needed when just EKAT_ENABLE_FPE=ON could allow both debug and optimized builds to check and then could be something set on all CI tests. Just a thought.....

mjs271:
From the auto tester for double/debug:

```
The following tests FAILED:
 60 - run_stand_modal_aero_calcsize_sub (NUMERICAL)
 61 - validate_stand_modal_aero_calcsize_sub (Failed)
396 - run_ccncalc_ts_1400 (NUMERICAL)
397 - validate_ccncalc_ts_1400 (Failed)
412 - run_stand_dropmixnuc_ts_1407 (NUMERICAL)
413 - validate_stand_dropmixnuc_ts_1407 (Failed)
414 - run_dropmixnuc_ts_1400 (NUMERICAL)
415 - validate_dropmixnuc_ts_1400 (Failed)
416 - run_dropmixnuc_ts_1401 (NUMERICAL)
417 - validate_dropmixnuc_ts_1401 (Failed)
418 - run_dropmixnuc_ts_1402 (NUMERICAL)
419 - validate_dropmixnuc_ts_1402 (Failed)
420 - run_dropmixnuc_ts_1403 (NUMERICAL)
421 - validate_dropmixnuc_ts_1403 (Failed)
422 - run_dropmixnuc_ts_1404 (NUMERICAL)
423 - validate_dropmixnuc_ts_1404 (Failed)
424 - run_dropmixnuc_ts_1405 (NUMERICAL)
425 - validate_dropmixnuc_ts_1405 (Failed)
426 - run_dropmixnuc_ts_1406 (NUMERICAL)
427 - validate_dropmixnuc_ts_1406 (Failed)
428 - run_dropmixnuc_ts_1407 (NUMERICAL)
429 - validate_dropmixnuc_ts_1407 (Failed)
430 - run_dropmixnuc_ts_1408 (NUMERICAL)
431 - validate_dropmixnuc_ts_1408 (Failed)
432 - run_dropmixnuc_ts_1409 (NUMERICAL)
433 - validate_dropmixnuc_ts_1409 (Failed)
434 - run_dropmixnuc_ts_1410 (NUMERICAL)
435 - validate_dropmixnuc_ts_1410 (Failed)
436 - run_dropmixnuc_ts_1411 (NUMERICAL)
437 - validate_dropmixnuc_ts_1411 (Failed)
438 - run_dropmixnuc_ts_1412 (NUMERICAL)
439 - validate_dropmixnuc_ts_1412 (Failed)
440 - run_dropmixnuc_ts_1413 (NUMERICAL)
441 - validate_dropmixnuc_ts_1413 (Failed)
442 - run_dropmixnuc_ts_1414 (NUMERICAL)
443 - validate_dropmixnuc_ts_1414 (Failed)
444 - run_dropmixnuc_ts_1415 (NUMERICAL)
445 - validate_dropmixnuc_ts_1415 (Failed)
446 - run_dropmixnuc_ts_1416 (NUMERICAL)
447 - validate_dropmixnuc_ts_1416 (Failed)
448 - run_dropmixnuc_ts_1417 (NUMERICAL)
449 - validate_dropmixnuc_ts_1417 (Failed)
```

... so at least we can replicate here 🙂

overfelt:
Interesting, the dropmixnuc failures are from numerical nan's in the input yaml files. Causes skywalker to throw an floating point exception.

jeff-cohere:
> Interesting, the dropmixnuc failures are from numerical nan's in the input yaml files. Causes skywalker to throw an floating point exception.

Hopefully it's not too hard to get rid of those!

mjs271:
@overfelt @jeff-cohere

In light of James' comment, I just pushed up a potential way for enabling FPEs at run time in validation testing, via overloading `validation::initialize()`.

I think this position for enabling FPEs at the top of the driver is safest, since we've seen that skywalker can error when it ingests bad numerics.

Another reason I'd favor this approach is that now `validation.hpp` contains the names of the `default_fpes`, so they're available for reference, and this doesn't require searching through ekat to try and remember the function/syntax. Alternatively, an approach with less machinery would be
```cpp
int main(int argc, char **argv) {
  if (argc == 1) {
    usage();
  }
  validation::initialize(argc, argv); // FPEs turned off--good
  ekat::enable_fpes(validation::default_fpes); // FPEs turned on--works
  ekat::enable_fpes(FE_OVERFLOW); // specific FPE turned on--works
  [...]
}
```
but that relies on knowing that ekat contains this capability. However, none of this prevents adding these calls to individual tests, functions, etc.

I am curious on other's thoughts as to whether FPEs should be enabled or disabled by default for tests, and whether to apply the same rule to both validation and unit tests. I'm in favor of on by default for all tests to ensure that we catch bugs that can appear invisible, but that decision could also have knock-on effects that I'm not thinking about.

odiazib:
> Thanks, Mike for catching this! @odiazib: would this also be able to detect the NaN's that were revealed in scream AT tests for the optics branch?

Hi @singhbalwinder,  I believe the AT relies on this EKAT's feature to detect FPE. Thus, yes, this would be able to detect NanN's that were reported by AT.

jeff-cohere:
@mjs271, can you paste the code you pushed into your prior comment for clarity, so we can see what's happening alongside your explanation, and compare/contrast it with your alternative?

mjs271:
@jeff-cohere Here you go:

This is what's in `validation.hpp`
```cpp
#include <cfenv>

namespace mam4 {

[...]

/// Call this function to initialize a validation driver.
void initialize(int argc, char **argv);

/// initialize with FPEs enabled, provided via argument
constexpr int default_fpes = FE_DIVBYZERO | FE_INVALID | FE_OVERFLOW;
void initialize(int argc, char **argv, const int fpes_);
```
`validation.cpp`
```cpp
namespace mam4 {
namespace validation {

void initialize(int argc, char **argv) {
  Kokkos::initialize(argc, argv);
}

void initialize(int argc, char **argv, int fpes_) {
  Kokkos::initialize(argc, argv);
  ekat::enable_fpes(fpes_);
}
```
example implementation in `calcsize_driver.cpp`
```cpp
int main(int argc, char **argv) {
  if (argc == 1) {
    usage();
  }
  // FPEs turned off
  // validation::initialize(argc, argv);
  // default FPEs turned on
  validation::initialize(argc, argv, validation::default_fpes);
```

jeff-cohere:
This looks good to me. What do others think?

overfelt:
Sure, merge it.

jeff-cohere (approved):

jeff-cohere:
I leave it to those involved to decide how best to proceed with merging. Perhaps it's least disruptive to fix the tests in this branch, but the integration activity is separate enough from mam4xx at this point that we could have broken tests in `main` if merging now seems more expedient.

src/validation/mam_x_validation
odiazib, line 1: @mjs271 , why is  this PR adding 25 new files in `mam_x_validation`?

odiazib:
> I leave it to those involved to decide how best to proceed with merging. Perhaps it's least disruptive to fix the tests in this branch, but the integration activity is separate enough from mam4xx at this point that we could have broken tests in `main` if merging now seems more expedient.

Hi @jeff-cohere, we should fix these tests before merging. I will check these fail test shortly.

src/validation/mam_x_validation
mjs271: Haha, good catch. I bumped `mam_x_validation` to the current `origin/main`, which includes James' PR that gets rid of NaNs, and some of the work I've done on `mo_drydep`.

And, as it turns out, it's up to 67 now 😂

odiazib:
With the later fixes from @mjs271  and @overfelt . Only one test is failing in my linux box. 
`39 - run_stand_modal_aero_calcsize_sub (NUMERICAL)`.

odiazib:
> Interesting, the dropmixnuc failures are from numerical nan's in the input yaml files. Causes skywalker to throw an floating point exception.

The `run_stand_modal_aero_calcsize_sub` failures are also because of input-yaml-files issue. In particular, the yaml file has this number `0.1797693135E+309` and Skywalker assigned an `inf`.

odiazib:
> With the later fixes from @mjs271 and @overfelt . Only one test is failing in my linux box. `39 - run_stand_modal_aero_calcsize_sub (NUMERICAL)`.

After @mjs271 updated `mam_x_validation` with the fixes from @overfelt  all tests are passing.

odiazib (approved):
It looks good to me. Thanks for adding FPE check.

singhbalwinder (approved):
Looks good to me too!

jeff-cohere:
Nice work!

mjs271:
Note that unit test FPEs are enabled via haero, and added in [haero/#479](https://github.com/eagles-project/haero/pull/479) (branch `haero/mjs/FPEs_4unitTests`).

overfelt:
I think all the tests are fixed for the CPU runs.  Just waiting for the pnnl GPU tests to run.

mjs271:
Ah, wow--thank you, @overfelt!

overfelt (approved):

overfelt (approved):

PR #306, opened 2024-03-22
Fix for reordering of aerosols within modes.

overfelt:
Balwinder has a branch in which he changes the ordering of aerosols within modes. This should not make a difference but an extra call to aerosol_index_for_mode was breaking Gas/Aerosol exchange in this case.  Once this fix is merged into Balwinder's branch, it should fix his tests.

### DIFF
--- a/src/mam4xx/mo_setsox.hpp
+++ b/src/mam4xx/mo_setsox.hpp
@@ -75,7 +75,7 @@ struct Config {
   int itermax = 20;
   // Real small_value_20 = 1.0e-20;
   // Real small_value_30 = 1.0e-30;
-  int numptrcw_amode[AeroConfig::num_modes()] = {23, 28, 36, 40};
+  int numptrcw_amode[AeroConfig::num_modes()] = {22, 27, 35, 39};
 
   Config() = default;
   Config(const Config &) = default;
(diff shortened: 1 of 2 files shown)


src/mam4xx/mo_setsox.hpp
```diff
@@ -75,7 +75,7 @@ struct Config {
   int itermax = 20;
   // Real small_value_20 = 1.0e-20;
   // Real small_value_30 = 1.0e-30;
-  int numptrcw_amode[AeroConfig::num_modes()] = {23, 28, 36, 40};
+  int numptrcw_amode[AeroConfig::num_modes()] = {22, 27, 35, 39};
```
singhbalwinder, line 78: Is this change related to my species order change? To me, it looks like a fix to accommodate C++'s zero-based array indexing.

singhbalwinder (approved):
Thanks for fixing this!

src/mam4xx/mo_setsox.hpp
overfelt: Yes, it has nothing to do with your species order change and is just a Fortran to C++ indexing problem. But the commit integration tests failed due to the memory error and i needed to fix it to pass all of the tests.

PR #308, opened 2024-03-25
Adds a utility function to get HAERO's prognostic array index

singhbalwinder:
In the EAMxx interface routine, we construct dry/wet aerosol arrays reading species from an ordered array. This function (originally written by James) maps the index of the species to HAERO's prognostic array.

### DIFF
--- a/src/mam4xx/utils.hpp
+++ b/src/mam4xx/utils.hpp
@@ -174,11 +174,19 @@ void transfer_work_arrays_to_prognostics(const Real q[gas_pcnst()],
 
 #undef DECLARE_PROG_TRANSFER_CONSTANTS
 
+// Get specie index of the HAERO's prognostics array given a specie index
+KOKKOS_INLINE_FUNCTION
+int get_haero_prognostics_index(const int mode, const int species_index) {
+  const mam4::AeroId aero_id = mam4::mode_aero_species(mode, species_index);
+  const int ind =
+      (mam4::AeroId::None != aero_id) ? static_cast<int>(aero_id) : -1;
+  return ind;
+};
+
 // Given an AerosolState with views for dry aerosol quantities, creates a
 // interstitial aerosols 1D view (state_q) for the column with the given index.
 // This object can be provided to mam4xx for the column.
 
-// MUST FIXME: address James comments about making the code better.
 KOKKOS_INLINE_FUNCTION
 void extract_stateq_from_prognostics(const mam4::Prognostics &progs,
                                      const haero::Atmosphere &atm, Real *q,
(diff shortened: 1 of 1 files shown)


overfelt (approved):

src/mam4xx/utils.hpp
```diff
@@ -209,7 +217,9 @@ void extract_stateq_from_prognostics(const mam4::Prognostics &progs,
   for (int m = 0; m < AeroConfig::num_modes(); ++m) {
     // First add the aerosol species mmr
     for (int a = 0; a < mam4::num_species_mode(m); ++a) {
-      q[s_idx] = progs.q_aero_i[m][a](klev);
+      const int h_ind = get_haero_prognostics_index(
```
odiazib, line 220: Why do you name this function `get_haero_prognostics_index `? is the prognostics struct defined in `mam4xx`. should we call this function `get_prognostics_index `?

src/mam4xx/utils.hpp
```diff
@@ -209,7 +217,9 @@ void extract_stateq_from_prognostics(const mam4::Prognostics &progs,
...
+          m, a); // Index of HAERO's prognostics array
+      q[s_idx] = progs.q_aero_i[m][h_ind](klev);
```
odiazib, line 222: The code will invoke `get_haero_prognostics_index` inside of levels (kk) and cols (icols) loop. We must check if this part of code becomes a bottleneck.

odiazib (changes requested):
Just a  minor change.

src/mam4xx/utils.hpp
singhbalwinder: I used `haero` in the name as `prognostics` is defined in `haero` (see [here](https://github.com/eagles-project/haero/blob/720bd29e8884eeba8dfcc9fc75476d5c6984ca85/haero/aero_process.hpp#L25))

I will change the name as adding `haero` doesn't add any new info the the function (we have just one `prognostics`)

src/mam4xx/utils.hpp
singhbalwinder: Is there an easy way to check it? To me, it looked more like an array look-up. Do you think it might cause noticeable slowdowns?

odiazib: I hope that @cameronrutherford will tells if this type of code causes slowdowns. For my understanding, he will carry out a performance analysis for mam4xx.

odiazib (approved):
It looks good to me.

src/mam4xx/utils.hpp
```diff
@@ -209,7 +217,9 @@ void extract_stateq_from_prognostics(const mam4::Prognostics &progs,
   for (int m = 0; m < AeroConfig::num_modes(); ++m) {
     // First add the aerosol species mmr
     for (int a = 0; a < mam4::num_species_mode(m); ++a) {
-      q[s_idx] = progs.q_aero_i[m][a](klev);
+      const int prg_idx =
```
odiazib, line 220: I reviewed this part of the code and found that we cannot use ` mam4::mode_aero_species(mode, species_index)` to obtain the index of the species in each mode.  ` mam4::mode_aero_species(mode, species_index)` returns the aero_id, which is an enum from [here](https://github.com/eagles-project/mam4xx/blob/7d14b8a71166c48327f1839c46c0f974134a631f/src/mam4xx/aero_modes.hpp#L130).

Because mam4xx prognostics structure has same ordering of species per node that state_q and qqcw, we do not need to add this function; however, we could add use the function that @jeff-cohere have implemented [here](https://github.com/eagles-project/mam4xx/blob/7d14b8a71166c48327f1839c46c0f974134a631f/src/mam4xx/aero_modes.hpp#L249) and it is used [here](https://github.com/eagles-project/mam4xx/blob/7d14b8a71166c48327f1839c46c0f974134a631f/src/mam4xx/utils.hpp#L163).

odiazib (changes requested):

singhbalwinder:
We no longer need this.

PR #318, opened 2024-05-09
aerosol_optics - switching for to Kokkos::parallel_fors.

odiazib:
We encountered a race condition in the CIME case involving the `SMS_D_Ln5.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.chrysalis_intel.scream-mam4xx-optics `(D is for debug) and the` eamxx-multi-process` test (`homme_shoc_cld_p3_mam_optics_rrtmgp`). This issue was reproducible on multiple machines with different compilers. The race condition only occurred with the debug build type and multiple threads. I discovered that switching the for loop to Kokkosparallel_fors resolves this race condition. As a precaution, I added a` Kokkos::single` over `tropopause_or_quit`, which is serial routine.

Switching the for loop to` Kokkos::parallel_for`s will improve performance. However, it is unclear why our eamxx test only failed with the debug build type and multiple threads on the host.

### DIFF
--- a/src/mam4xx/aer_rad_props.hpp
+++ b/src/mam4xx/aer_rad_props.hpp
@@ -37,8 +37,8 @@ constexpr Real cnst_ka1 = cnst_kap - 1.0;
 // Similar to volcanic_cmip_sw,
 // But, tau, tau_w, tau_w_g, tau_w_f have layout (nswbands, pver)
 KOKKOS_INLINE_FUNCTION
-void volcanic_cmip_sw2(const ConstColumnView &zi, const int ilev_tropp,
-                       const View2D &ext_cmip6_sw_inv_m,
+void volcanic_cmip_sw2(const ThreadTeam &team, const ConstColumnView &zi,
+                       const int ilev_tropp, const View2D &ext_cmip6_sw_inv_m,
                        const View2D &ssa_cmip6_sw, const View2D &af_cmip6_sw,
                        const View2D &tau, const View2D &tau_w,
                        const View2D &tau_w_g, const View2D &tau_w_f) {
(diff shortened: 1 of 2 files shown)


singhbalwinder (approved):
Thanks for fixing this weird issue.

jeff-cohere:
Race conditions are not always reproducible, so I'm not too concerned about it not manifesting in all cases. Thanks for fixing this, Oscar.

jeff-cohere (approved):

ambrad:
Just to clarify some things in case this helps:

1. The `parallel_for` change indeed fixes a clear _write_ race condition. For example, a line that updates a variable based on its current value, like `tau = ... tau ...`, is a write race condition if multiple threads are writing that variable (at the same index) simultaneously. Here, the key is that a written variable is also read.

2. The reason you see it in debug is because the compiler does a literal translation of the code. In optimized builds, often the inner loop is implemented in a nonstraightforward way, e.g., by introducing temporaries, that can hide a write race condition. For example, imagine the compiler turns the RHS variables into temporaries before any of the writes start. This can make the race condition happen only very rarely or never. This issue of race conditions being exhibited in debug vs opt builds comes up fairly frequently, including in the mainstream E3SM F90 code.

3. The `single`s are, as you probably suspected, not needed. This is a case of a `benign` write race condition. A simple example is `x(i) = 1`. It's OK is many threads execute that line for the same `i` value because the RHS doesn't depend on data in flight.

singhbalwinder:
Thanks, @ambrad for the very nice explanation. I was very confused about it, but your insights helped clear that confusion. I always thought the `for` loop was safer than `parallel_for` for avoiding a race condition as a `for` loop is serial. Here we saw the opposite. Does `parallel_for` introduce compiler-like optimization or does it help because it creates temporary private variables?

ambrad:
@singhbalwinder these `parallel_for` loops are _inside_ a kernel. When the loop is inside a kernel, you have to be careful about making each thread do the right thing. Kokkos's inner `parallel_for`s are designed to do just that. A bare `for` makes each thread run the whole inner loop, creating read/write conflicts in the cases that this PR fixes. An inner `parallel_for` instead assigns one thread per index value.

PR #325, opened 2024-06-04
MAM4xx changes needed for the ACI process

singhbalwinder:
While working on the ACI codes, I need to make several changes to the MAM4xx. Most of these changes are related to indexing. I also swapped the dimensions of `factnum` so that I could slice it using `subview`.

### DIFF
--- a/src/mam4xx/hetfrz.hpp
+++ b/src/mam4xx/hetfrz.hpp
@@ -1234,25 +1234,48 @@ void hetfrz_rates_1box(const int k, const Real dt, const Atmosphere &atm,
   const int accum_idx = int(ModeIndex::Accumulation);
   const int pcarbon_idx = int(ModeIndex::PrimaryCarbon);
 
-  auto &accum_bc  = progs.q_aero_i[accum_idx][aerosol_index_for_mode(ModeIndex::Accumulation, AeroId::BC)];
-  auto &accum_dst = progs.q_aero_i[accum_idx][aerosol_index_for_mode(ModeIndex::Accumulation, AeroId::DST)];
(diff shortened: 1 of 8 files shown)


singhbalwinder:
This PR might break so tests as I changed the indices. We might need to make changes to the validation test drivers.

src/mam4xx/hetfrz.hpp
```diff
@@ -1234,25 +1234,25 @@ void hetfrz_rates_1box(const int k, const Real dt, const Atmosphere &atm,
   const int accum_idx = int(ModeIndex::Accumulation);
   const int pcarbon_idx = int(ModeIndex::PrimaryCarbon);
 
-  auto &accum_bc = progs.q_aero_i[accum_idx][int(AeroId::BC)];
-  auto &accum_dst = progs.q_aero_i[accum_idx][int(AeroId::DST)];
-  auto &accum_soa = progs.q_aero_i[accum_idx][int(AeroId::SOA)];
-  auto &accum_so4 = progs.q_aero_i[accum_idx][int(AeroId::SO4)];
-  auto &accum_pom = progs.q_aero_i[accum_idx][int(AeroId::POM)];
-  auto &accum_mom = progs.q_aero_i[accum_idx][int(AeroId::MOM)];
+  auto &accum_bc  = progs.q_aero_i[accum_idx][aerosol_index_for_mode(ModeIndex::Accumulation, AeroId::BC)];
```
odiazib, line 1237: This change is needed. However, I do not like the fact that we need to use ` ModeIndex::Accumulation` twice in progs. Should we create an issue to improve progs?

odiazib (approved):

src/mam4xx/hetfrz.hpp
singhbalwinder: I agree, we can look into it to see how to fix this. I will create an issue.

PR #328, opened 2024-06-11
Fix warnings due to signed/unsigned integer comparison

bartgol:
Steps towards a warning free eamxx build.

### DIFF
--- a/src/mam4xx/convproc.hpp
+++ b/src/mam4xx/convproc.hpp
@@ -3026,8 +3026,8 @@ void ma_convproc_intr(
   auto qnew = Diagnostics::ColumnTracerView(
       scratch1Dviews[ConvProc::Col1DViewInd::qnew].data(), nlev,
       aero_model::pcnst);
-  EKAT_KERNEL_ASSERT(state_q.extent(0) == nlev);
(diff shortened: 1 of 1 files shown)


mjs271 (approved):

bartgol:
@mjs271 @odiazib @jeff-cohere I'm not familiar with the gh actions for mam4xx. The codecov fail seems to be due to a submit error, while the other is a git push fail that usually happens when the remote is not configured correctly. Maybe the gh action assumes the PR is NOT from a fork? Asking for a friend...

jeff-cohere:
> @mjs271 @odiazib @jeff-cohere I'm not familiar with the gh actions for mam4xx. The codecov fail seems to be due to a submit error, while the other is a git push fail that usually happens when the remote is not configured correctly. Maybe the gh action assumes the PR is NOT from a fork? Asking for a friend...

I believe we've seen exactly this before on PRs from forks. I don't think we've looked into making CodeCov work properly with forks, but it might be time to take a closer look at that.

In any case, I think we can merge this. Let me know if you want me to do so.

bartgol:
Yes, let's merge. I am updating mam4xx in the scream repo to get shared libs. I can re-update to get this fix in if we merge it.

PR #331, opened 2024-07-02
Update CMakeLists.txt for Validation Diffs Due to Relative Error Change

mjs271:
This PR is in response to the [mam_x_validation PR #72](https://github.com/eagles-project/mam_x_validation/pull/72) that changed validation testing to always use relative error as the pass/fail criteria.

Unfortunately, this led to some reduced tolerances ~~and some failing tests due to `NaN` values~~. In response, I made edits to the relevant `CMakeLists.txt` files to result in all enabled tests passing. I have denoted all changes with `FIXME` comments that mention the change to relative error criteria. Most tests lost an order of magnitude or two in accuracy, which may not necessarily be a dire issue, though a few ended up with relative errors that were quite large, e.g., $\geq \mathcal{O}(1)$. ~~However, the tests with hidden `NaN` values will likely require attention, and they are currently disabled.~~

_**Update:**_ The `NaN`'s were a result of my mistake--caused by zero-error that needed to be checked for and is fixed with the latest push.

### DIFF
--- a/src/validation/aero_model/CMakeLists.txt
+++ b/src/validation/aero_model/CMakeLists.txt
@@ -39,9 +39,9 @@ set(TEST_LIST
 set(DEFAULT_TOL 1e-11)
 set(DEFAULT_TOL_BASELINE 1e-13)
 
-set(ERROR_THRESHOLDS  
-   5e-8
-   ${DEFAULT_TOL}
+set(ERROR_THRESHOLDS
+   3e-6  # calc_1_impact_rate_ts_0--FIXME?--tol changed in repsonse to using rel error for pass/fail
+   8e-10 # modal_aero_bcscavcoef_get_ts_355--FIXME?--tol changed in repsonse to using rel error for pass/fail
    2e-4
    2e-5
    2e-5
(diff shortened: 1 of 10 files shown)


jeff-cohere:
Can we definitely state that we should be using relative errors to judge all test results? In my experience this stuff often requires some judgement based on the quantities themselves. I'm curious what the reviewers of [thje related mam_x_validation PR](https://github.com/eagles-project/mam_x_validation/pull/72) think of this.

mjs271:
> Can we definitely state that we should be using relative errors to judge all test results? In my experience this stuff often requires some judgement based on the quantities themselves. I'm curious what the reviewers of [thje related mam_x_validation PR](https://github.com/eagles-project/mam_x_validation/pull/72) think of this.

There's definitely room for debate--this was just what made sense to me. I was surprised by how many tests have _**large**_ relative errors, like $\mathcal{O}(1)$ and greater, but were still passing based on the previous criteria.

Upon thinking on it a little more, I suppose there could be cases where the relative error could be skewed by entries in an output array that vary many orders of magnitude (I think?). However, I don't yet have well-formed ideas on how to handle that case

PR #332, opened 2024-07-04
Adds comments to describe I/O and makes a few variables constants to match EAMxx datatypes

singhbalwinder:
While working on the wet scavenging interface, I came across several variables that were inputs for the process. All input variables should be declared as constants in EAMxx. I had to change those variables to `const`s in wet_dep.hpp as well.


src/mam4xx/wet_dep.hpp
```diff
@@ -2061,6 +2114,7 @@ class WetDeposition {
 
   const char *name() const { return "MAM4 Wet Deposition"; }
 
+  KOKKOS_INLINE_FUNCTION
```
singhbalwinder, line 2117: It should be just an `inline` function as it is called from the host, right?

overfelt: Yes, "init" is generally not callable from device since it can allocate memory.

odiazib (approved):
It looks good to me.

PR #334, opened 2024-07-16
Use TeamVectorRange policies

tcclevenger:
Changes:
- Use `TeamVectorRange` if only a single level of parallelism exists within a team policy loop. Given the TeamPolicy used in EAMxx for MAM4xx code, I don't think these fixes will solve non-determinism, but if in the future a vector length >1 is used in the TeamPolicy it might be necessary.
- Use `[&]` for the capture in inner parallel loops. This may clear up some compiler errors/warnings on some arch and should be slightly more performant.
- Found a case where a parallel_reduce had the local and global sum variables having the same name. This should work fine, but is confusing.


singhbalwinder:
@odiazib, @mjs271 , @overfelt : I have tested it on the PNNL deception machine and all the tests are passing. Should I go ahead and merge it?

odiazib (approved):

PR #337, opened 2024-07-31
Adds an array of molecular weights needed for surface emissions unit conversion

singhbalwinder:
The molecular weight array was present in validation unit test files. This PR copies it to the source code as well so that we can use it in EAMxx interfaces.


singhbalwinder:
Didn't realize that it has other codes as well. Closing it.

Issue #341, opened 2024-08-06
Create unit tests for `aero_model_emissions`


mjs271:
`aero_model_emissions` could probably use a few more tests, so I plan to revisit and write some unit tests.

Issue #340, opened 2024-08-06
Finish remaining validation tests for `aero_model_emissions`


mjs271:
As of PR submission, 2 tests require extra data:
- `marine_organic_emis()`
- `marine_organic_numflx_calc()`

Upon getting the necessary variables, I expect the tests to pass readily.

PR #342, opened 2024-08-06
Aerosol Model Emissions Port

mjs271:
This PR ports the vertical (online) aerosol emissions from the `aerosol_model`. The 3 top-level functions involve updating:
- seasalt emissions
- marine organic emissions (technically a subset, or at least intersection with seasalt)
- dust emissions

All enabled validation tests are passing. However, 2 tests require more data as of right now--namely:
- `marine_organic_emis()`
- `marine_organic_numflx_calc()`
and this is noted in Issue #340.

Additionally, I note in Issue #341 that some unit tests are probably in order, and I will revisit that.


overfelt:
The aerosol modal emissions functions run on the GPU but the validation tests are CPU only.  Maybe run the tests on the GPU?

src/mam4xx/aero_model_emissions.hpp
```diff
@@ -0,0 +1,919 @@
...
+  // [m3 m-3] Volume concentration resolved
+  Real vlm_rsl[nbin] = {0.0};
+
+  // Compute resolved size statistics for each size distribution
+  // In C. Zender's code call dst_sz_rsl
+  // Factor for logarithmic grid
+  Real series_ratio;
+  for (int n = 0; n < nbin; ++n) {
+    series_ratio = haero::pow((dmt_max[n] / dmt_min[n]), (1.0 / sz_nbr));
+    sz_min[1] = dmt_min[n];
```
overfelt, line 259: Fortran indexing mistake? I think `sz_min[1]` should be `sz_min[0]`.

src/mam4xx/aero_model_emissions.hpp
```diff
@@ -0,0 +1,919 @@
...
+    // NOTE: Loop starts at 1 (2 in fortran code)
+    for (int m = 1; m < sz_nbr; ++m) {
+      sz_min[m] = sz_min[m - 1] * series_ratio;
+    }
+    // Derived grid values
+    // NOTE: Loop ends at sz_nbr - 2 (fortran: sz_nbr - 1)
+    for (int m = 0; m < sz_nbr - 2; ++m) {
+      sz_max[m] = sz_min[m + 1];
+    }
+    sz_max[sz_nbr] = dmt_max[n];
```
overfelt, line 269: Index out-of-bounds: `sz_max[sz_nbr] `should be `sz_max[sz_nbr-1]`.

mjs271: Clearly transcribing 1-based to 0-based languages is my kryptonite... 🙃

If there's one thing I can hope to accomplish in my career, it'd be to develop those **_J. Overfelt Eagle Eyes_** ™️ 🙂

Thank you!

mjs271:
@odiazib @singhbalwinder @overfelt @jaelynlitz 

All tests now passing--should be good to go! 🚀

singhbalwinder:
Hi @mjs271 : Perhaps I am missing something but I don't see `dust_dmt_vwr` computed anywhere in the ported code. [Here](https://github.com/eagles-project/e3sm_mam4_refactor/blob/aab9c07c85d852a1fb552f119a47b40525433ccf/components/eam/src/chemistry/aerosol/dust_common.F90#L30C46-L30C53) is the fortran function that computes it. I can see the code but it is all commented out.

singhbalwinder (changes requested):
Nice work! I like how you improved the code at some places while porting it. I have left very minor comments.

src/mam4xx/aero_model_emissions.hpp
```diff
@@ -0,0 +1,972 @@
...
+
+#include <mam4xx/aero_config.hpp>
+#include <mam4xx/gas_chem_mechanism.hpp>
+#include <mam4xx/mam4_types.hpp>
+#include <mam4xx/utils.hpp>
+
+namespace mam4::aero_model_emissions {
+
+constexpr int pcnst = mam4::pcnst;
+constexpr int n_online_emiss = 9;
```
odiazib, line 14: Should we add BAD CONSTANT?

src/mam4xx/aero_model_emissions.hpp
```diff
@@ -0,0 +1,972 @@
...
+
+enum class FluxType { MassFlux, NumberFlux };
+
+// =============================================================================
+//      Dust Parameters/Constants
+// =============================================================================
+// nbin corresponds to dst_a1, dst_a3
+const int dust_nbin = 2;
```
odiazib, line 22: Should we add BAD CONSTANT?

src/mam4xx/aero_model_emissions.hpp
```diff
@@ -0,0 +1,972 @@
...
+// nnum corresponds to num_a1, num_a3
+const int dust_nnum = 2;
```
odiazib, line 24: Should we add BAD CONSTANT?

odiazib: The same applies to all the constant variables in this code

src/mam4xx/aero_model_emissions.hpp
```diff
@@ -0,0 +1,972 @@
...
+    // Mass weighted diameter resolved [m]
+    dust_dmt_vwr[n] /= vlm_rsl[n];
+  } // end for (dust_nbin)
+} // end init_dust_dmt_vwr()
+
+KOKKOS_INLINE_FUNCTION
+void dust_emis(
+    // in
+    const int dust_indices[dust_nbin + dust_nnum], const Real dust_density,
+    const Real (&dust_flux_in)[dust_nflux_in], const DustEmissionsData data,
```
odiazib, line 285: @mjs271 , Can you explain to me what this input does: `const Real (&dust_flux_in)[dust_nflux_in]`? What is different from the input `const Real dust_flux_in[dust_nflux_in]`?.

odiazib (changes requested):
It looks great! I just added a few comments to mark BAD CONSTANT.

src/mam4xx/aero_model_emissions.hpp
mjs271: As I understand it to work, `Real (&var)[var_size]` passes a reference to the array of `Real`s. This is in contrast to the version without parentheses, `Real &var[var_size]` that passes an array of references to `Real`s and is [categorically illegal in C++](https://stackoverflow.com/questions/1164266/why-are-arrays-of-references-illegal).

From a practical standpoint, I was trying to do the illegal one and ran across this method for making it work. Whether passing by reference is even worth it here (or a good idea)... well that's another question 🙃

All that being said, I'm happy to change it if it's dumb or causes confusion

src/mam4xx/aero_model_emissions.hpp
mjs271: I was wondering this, myself, and my thought was that the number and types of tracers in the current implementation of MAM4xx is not configurable, so these hard-coded `int`s are more like fixed model parameters that are only related to online emissions. However, when the day comes that we want to generalize the model, it'd probably be useful to see where numbers are hard-coded.

I'll add in the BAD CONSTANT's , for safety's sake, but am also interested in others' thoughts on the above

odiazib: Should we use another label/comment, for instance, HARD-CODED? @mjs271  @singhbalwinder. I agree that adding this type of comment will help us with moving these parameters to a different place.

mjs271: I do like that idea. For now, I decided to relegate the entire top section to a `BAD_CONSTANTS` namespace since that seemed easier than writing it on every line 😁

Happy to change it to whatever seems best to the group

singhbalwinder: Where is `n_online_emiss` used? That will help us find how it is used.

mjs271: ... unused and deleted 🙄

I think it may have been in there from my initial crack at the interface.

mjs271:
@odiazib @singhbalwinder 

Anyone opposed to merging? If I don't hear from you, I'll plan to merge early on Wed 8/28

singhbalwinder (approved):

odiazib (approved):

PR #343, opened 2024-09-12
Oscar/soaexch

odiazib:
validation test and port updates for:
* `mam_soaexch_1subarea`
* `mam_gasaerexch_1subarea`
* `gas_aer_uptkrates_1box1gas`


src/mam4xx/gasaerexch.hpp
```diff
@@ -319,6 +319,91 @@ Real fuchs_sutugin(const Real &D_p, const Real &gasfreepath,
   return fuchs_sutugin;
 }
 
+KOKKOS_INLINE_FUNCTION
+void gas_aer_uptkrates_1box1gas_OD(const Real accom, const Real gasdiffus,
```
mingxuanwupnnl, line 323: When it is ready, we may want to remove OD for the new function name and the old function as well.

src/mam4xx/gasaerexch.hpp
```diff
@@ -863,6 +946,208 @@ void GasAerExch::compute_tendencies(const AeroConfig &config,
         alnsg_aer);
   });
 }
+
+KOKKOS_INLINE_FUNCTION
+void mam_gasaerexch_1subarea(
```
mingxuanwupnnl, line 951: when ready, we may need to remove the old function.

mingxuanwupnnl (changes requested):

src/mam4xx/gasaerexch_soaexch.hpp
```diff
@@ -61,6 +61,259 @@ void soa_equilib_mixing_ratio_no_solute(const Real &T_in_K,     // in
...
+  Real delh_vap_soa[ntot_soaspec] = {156.0e3};
+  Real del_g_soa_tmp[ntot_soaspec] = {0.0};
+  Real g0_soa[ntot_soaspec] = {0.0};
+  Real g_soa[ntot_soaspec] = {0.0};
+  Real g_star[ntot_soaspec][max_mode] = {{0.0}};
+  Real opoa_frac[ntot_poaspec][max_mode] = {{0.1}};
+  Real phi[ntot_soaspec][max_mode] = {{0.0}};
+  Real p0_soa[ntot_soaspec] = {1.0e-10};
+  Real p0_soa_298[ntot_soaspec] = {1.0e-10};
+  Real sat[ntot_soaspec][max_mode] = {{0.0}};
```
singhbalwinder, line 119: You probably already know, but I found out in the past that initializing arrays like this doesn't actually initialize all array elements. C++ initializes the first element to the value in the curly brackets and the rest are initialized to zero. It is not guaranteed to work on GPU.

src/mam4xx/gasaerexch.hpp
odiazib: I will delete OD. We can keep both versions of this routine for now.

src/mam4xx/gasaerexch.hpp
odiazib: We should plan to delete part of the code in a future pull request (PR).

src/mam4xx/gasaerexch.hpp
```diff
@@ -863,6 +946,208 @@ void GasAerExch::compute_tendencies(const AeroConfig &config,
...
+      const Real gas_diffus_igas = mam4::gasaerexch::gas_diffusivity(
+          temp, p_in_atm, mw_gas[igas], mw_air_gmol, vol_molar_gas[igas],
+          vol_molar_air);
+
+      // tmpb = mean_molecular_speed(temp, mw_gas[igas]);
+      // gas mean free path (m)
+      const Real molecular_speed = mam4::gasaerexch::mean_molecular_speed(
+          temp, mw_gas[igas], r_universal_mJ, r_pi);
+
+      const Real gas_freepath_igas = 3.0 * gas_diffus_igas / molecular_speed;
```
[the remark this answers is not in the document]
odiazib: Because this operation is a multiplication, I do not think it will make a significant difference whether we use non-decimal or decimal.

mingxuanwupnnl (approved):

src/mam4xx/gasaerexch.hpp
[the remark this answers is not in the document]
jeff-cohere: It makes no difference as far as I can tell. There is no way to multiply different arithmetic types in C++ without implicit casting taking place. The rules for implicit casting are [here](https://stackoverflow.com/questions/5563000/implicit-type-conversion-rules-in-c-operators).

src/mam4xx/gasaerexch_soaexch.hpp
odiazib: Thanks for catching this. It has been fixed in the code.

singhbalwinder (approved):

singhbalwinder:
Please feel free to merge it if you are not planning on adding more to this PR.

PR #344, opened 2024-09-12
Fix Failing Validation Tests for Online Emissions

mjs271:
I mistakenly pushed some updates to `mam_x_validation` that correspond to some working changes in the eamxx integration for `aero_model_emissions`. This adds the new changes to fix the failing tests on main.

### DIFF
--- a/src/mam4xx/aero_model_emissions.hpp
+++ b/src/mam4xx/aero_model_emissions.hpp
@@ -1,14 +1,14 @@
 #ifndef MAM4XX_AERO_MODEL_EMISSIONS_HPP
 #define MAM4XX_AERO_MODEL_EMISSIONS_HPP
 
-#include <haero/atmosphere.hpp>
-
 #include <mam4xx/aero_config.hpp>
 #include <mam4xx/gas_chem_mechanism.hpp>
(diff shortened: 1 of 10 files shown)


odiazib (approved):

singhbalwinder (approved):

mjs271:
To add some context:

This PR fixes some regressions for `aero_model_emissions` tests and also corrects/improves the code in a few minor ways.

- It turns out that the extent of the `cflux` array is `pcnst`, which is not the same as `gas_pcnst`.
  - That led to some propagating changes throughout `aero_model_emissions` and associated tests.
- I also changed a couple of validation tests to compare the full `cflux` output array (in contrast to only the elements that are operated on).
    - This was possible because, in those cases, the output values of `cflux` for the preceding function could be used as inputs.
    - Note that this doesn't make the current test any stronger but could potentially make things easier in the future.
- Because some of these modifications were done during the eamxx integration work, I also added some purely mechanical changes to the signature of the top-level `aero_model_emissions()` function.
  - E.g., passing fewer arguments, passing a view rather than an array, and moving some values to `data` structs.

PR #351, opened 2024-09-27
Add function to return reasonable upper and lower bounds of field values.

overfelt:


src/mam4xx/physical_limits.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+namespace mam4 {
+
+#if 0
+  // If a device callable version of physical_min and physical_max is needed along
+  // with the CPU only version, somthing like this would work. Strings do not do
+  // well on device but enums are fast.
+  
+  enum FieldNameIndex { T_mid, qv, qc, qi, nc, nr, ni, nmr, mmr, NUMFIELD};
+  struct min_max {const Real min; const Real max;};
```
jeff-cohere, line 20: Might this be better-named `MinMax`?

PR #355, opened 2024-10-03
Amicphys changes after BFB comparison with EAMv2

singhbalwinder:
Changes to several routines after EAMv2 comparison.


singhbalwinder:
I removed codes, and we have tests for those codes that are now failing. Since those codes are not used anymore, is it okay to get rid of the tests? I can create an issue for it so that we have tests for the new codes.

src/mam4xx/aero_modes.hpp
```diff
@@ -94,7 +94,7 @@ static constexpr Real mam4_coarse_mead_std_dev = 1.8;
 static constexpr Real mam4_primary_carbon_min_diameter_m = 1e-8;
 static constexpr Real mam4_primary_carbon_nom_diameter_m = 5e-8;
 static constexpr Real mam4_primary_carbon_max_diameter_m = 1e-7;
-static constexpr Real mam4_primary_carbon_mead_std_dev = 1.6;
+static constexpr Real mam4_primary_carbon_mead_std_dev = 1.60000002384186;
```
odiazib, line 97: I wonder if the last digit in these numbers is noise and if they are machine-dependent.

src/mam4xx/gasaerexch.hpp
```diff
@@ -33,7 +33,8 @@ class GasAerExch {
   static constexpr int num_aer = AeroConfig::num_aerosol_ids();
   static constexpr int nait = static_cast<int>(ModeIndex::Aitken);
   static constexpr int npca = static_cast<int>(ModeIndex::PrimaryCarbon);
-  static constexpr int igas_h2so4 = static_cast<int>(GasId::H2SO4);
+  static constexpr int igas_h2so4 = 1;
```
odiazib, line 36: Should we modify `GasId`? The current implementation has `GasId::H2SO4` as 2. I would change this enum class to reflect this change.

```
// Identifiers for gas species in MAM4
enum class GasId {
  O3 = 0,    // ozone
  H2O2 = 1,  // hydrogen peroxide
  H2SO4 = 2, // sulfuric acid
  SO2 = 3,   // sulfur dioxide
  DMS = 4,   // dimethyl sulfide
  SOAG = 5,  // secondary organic aerosol precursor
  None = 6,  // invalid gas id
};
```
These hard-coded indices can become a problem for us later when we modify the gas chemistry.

odiazib (changes requested):

odiazib:
> I removed codes, and we have tests for those codes that are now failing. Since those codes are not used anymore, is it okay to get rid of the tests? I can create an issue for it so that we have tests for the new codes.

Yes, we should remove these tests.

src/mam4xx/aero_modes.hpp
singhbalwinder: It is not machine-dependent. This number is picked up from the netcdf file that reads the modal data (see `sigmag` below from the netcdf file `mam4_mode4_rrtmg_c130628.nc`):
```
sigmag = 1.60000002384186 ;

 opticsmethod = "modal                           " ;

 dgnum = 5.00000005843049e-08 ;

 dgnumlo = 9.99999993922529e-09 ;

 dgnumhi = 1.0000000116861e-07 ;

 rhcrystal = 0.35 ;

 rhdeliques = 0.8 ;
```

It may be okay to let go of the last digits, but keeping them can help with the validation testing and also finding other discrepancies.

src/mam4xx/gasaerexch.hpp
singhbalwinder: Let me verify that. Amicphys has its own indexing and it uses only two gases. We may want to use this special indexing only for amicphys. 

I agree that this hard coding is problematic. We will address it in our code enhancement tasks next year.

singhbalwinder: Gas aerosol exchange uses just two gases: H2SO4 and SOAG. I have now removed all the unused indices in this file. We do not need to change anything else as these index values are specific to amicphys only.

mingxuanwupnnl (approved):

overfelt (approved):

PR #359, opened 2024-10-12
Changes return type of linoz srf sink function from void to Real

singhbalwinder:
The return type is changed so that the function's output is easy to see.

### DIFF
--- a/src/mam4xx/lin_strat_chem.hpp
+++ b/src/mam4xx/lin_strat_chem.hpp
@@ -259,7 +259,7 @@ void lin_strat_chem_solve(
 } // lin_strat_chem_solve
 
 KOKKOS_INLINE_FUNCTION
-void lin_strat_sfcsink_kk(const Real delta_t, const Real pdel, // in
+Real lin_strat_sfcsink_kk(const Real delta_t, const Real pdel, // in
                           Real &o3l_vmr, const Real o3_sfc, const Real o3_tau,
                           Real &do3mass) {
   constexpr Real one = 1.0;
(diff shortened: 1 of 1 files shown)


src/mam4xx/lin_strat_chem.hpp
```diff
@@ -259,7 +259,7 @@ void lin_strat_chem_solve(
 } // lin_strat_chem_solve
 
 KOKKOS_INLINE_FUNCTION
-void lin_strat_sfcsink_kk(const Real delta_t, const Real pdel, // in
+Real lin_strat_sfcsink_kk(const Real delta_t, const Real pdel, // in
                           Real &o3l_vmr, const Real o3_sfc, const Real o3_tau,
```
odiazib, line 263: would you need to remove `o3l_vmr` from the function signature?

odiazib (changes requested):

src/mam4xx/lin_strat_chem.hpp
singhbalwinder: It is both input and output, so we need to keep both.

src/mam4xx/lin_strat_chem.hpp
```diff
@@ -273,9 +273,9 @@ Real lin_strat_sfcsink_kk(const Real delta_t, const Real pdel, // in
       (one - haero::exp(-delta_t / o3_tau));     // !compute time scale factor
                                                  //
   const Real mass = pdel * rgrav;                //   air mass in kg/m2
-  const Real o3l_old = o3l_vmr;                  // vmr
+  const Real o3l_old = o3l_vmr_in;                  // vmr
   const Real do3 = (o3_sfc - o3l_old) * efactor; // vmr
-  o3l_vmr = o3l_old + do3;
+  const Real o3l_vmr = o3l_old + do3;
```
overfelt, line 278: With this change it does not appear that  `o3l_vmr_in` is an input/output value but input only. Maybe now it should be changed to `const Real.`

singhbalwinder: Thanks. I addressed this comment.

odiazib (approved):

mjs271 (approved):

PR #361, opened 2024-10-18
Implementing the find_season_index routine

odiazib:
Implementing the find_season_index routine to encapsulate the interpolation of the season_wes.nc file, from [line 267](https://github.com/eagles-project/e3sm_mam4_refactor/blob/599e772e34a0a32dacadc40688fd091e1f532d76/components/eam/src/chemistry/mozart/mo_drydep.F90#L267) to  [line 318](https://github.com/eagles-project/e3sm_mam4_refactor/blob/599e772e34a0a32dacadc40688fd091e1f532d76/components/eam/src/chemistry/mozart/mo_drydep.F90#L318).

This function is only invoked during initialization and has a small kernel. Thus, only a host version was implemented.

One validation test is implemented, and it is passing with a relative error tolerance of 0. Note that the output of the function (`index_season_lai`) consists of integers. Thus, an error of 0 is required to pass this test.

### DIFF
--- a/src/mam4xx/mo_drydep.hpp
+++ b/src/mam4xx/mo_drydep.hpp
@@ -6,6 +6,7 @@
 #ifndef MAM4XX_MO_DRYDEP_HPP
 #define MAM4XX_MO_DRYDEP_HPP
 
+#include <ekat/kokkos/ekat_subview_utils.hpp>
 #include <haero/math.hpp>
 #include <mam4xx/aero_config.hpp>
 #include <mam4xx/gas_chem_mechanism.hpp>
@@ -43,10 +44,14 @@ constexpr int nddvels = mam4::seq_drydep::n_drydep;
 
 // find_season_index only needs to be executed one time and is small.
 // Thus, execute it only on the host.
-
 using View1DHost = DeviceType::view_1d<Real>::HostMirror;
+using ConstView1DHost = DeviceType::view_1d<const Real>::HostMirror;
 using View1DIntHost = DeviceType::view_1d<int>::HostMirror;
+using View2DIntHost = DeviceType::view_2d<int>::HostMirror;
 using View3DIntHost = DeviceType::view_3d<int>::HostMirror;
+
+using KTH = ekat::KokkosTypes<ekat::HostDevice>;
+
 inline void find_season_index(const Real clat_j, const View1DHost &lat_lai,
                               const int nlat_lai, const View3DIntHost &wk_lai,
                               const View1DIntHost &index_season_lai) {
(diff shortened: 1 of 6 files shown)


singhbalwinder (approved):
Looks great!

src/mam4xx/mo_drydep.hpp
```diff
@@ -25,10 +26,111 @@ constexpr Real rair = 287.04;
 constexpr Real grav = 9.81;
 constexpr Real karman = 0.4;   // from shr_const_mod.F90
 constexpr Real tmelt = 273.15; // from shr_const_mod.F90 via physconst.F90
-
+constexpr Real r2d = 180.0 / haero::Constants::pi; // degrees to radians
```
jeff-cohere, line 29: This multiplicative factor converts radians to degrees, right? The variable name seems correct, but the comment is not.

src/mam4xx/mo_drydep.hpp
```diff
@@ -25,10 +26,111 @@ constexpr Real rair = 287.04;
...
+// Thus, execute it only on the host.
+using View1DHost = DeviceType::view_1d<Real>::HostMirror;
+using ConstView1DHost = DeviceType::view_1d<const Real>::HostMirror;
+using View1DIntHost = DeviceType::view_1d<int>::HostMirror;
+using View2DIntHost = DeviceType::view_2d<int>::HostMirror;
+using View3DIntHost = DeviceType::view_3d<int>::HostMirror;
+
+using KTH = ekat::KokkosTypes<ekat::HostDevice>;
+
+inline void find_season_index_at_lat(const Real clat_j,
```
jeff-cohere, line 55: Does this work without a KOKKOS_INLINE_FUNCTION  annotation? It's called in a `parallel_for` below.

jeff-cohere (approved):
Left a few comments. Looks good to me overall.

src/mam4xx/mo_drydep.hpp
odiazib: Yes, it should convert degrees to radians. It was a copy/paste error from [here]( https://github.com/eagles-project/e3sm_mam4_refactor/blob/599e772e34a0a32dacadc40688fd091e1f532d76/components/eam/src/chemistry/utils/mo_constants.F90#L32).

src/mam4xx/mo_drydep.hpp
odiazib: Yes, it does, because this find_season_index_at_lat is only invoked from the host

src/mam4xx/mo_drydep.hpp
jeff-cohere: But multiplying by 180.0 / pi converts radians to degrees, doesn't it?

2*pi * 180/pi = 2 * 180 = 360

singhbalwinder (approved):
Looks good! Thanks @odiazib !

src/mam4xx/mo_drydep.hpp
jeff-cohere: Ah, I see the KTH now. Thanks!

src/mam4xx/mo_drydep.hpp
odiazib: Yes. It should be radians to degrees.

odiazib: It is fixed in `mam4xx`.

PR #369, opened 2024-10-24
Always use TeamVectorRange and never TeamThreadRange.

overfelt:


src/mam4xx/aer_rad_props.hpp
```diff
@@ -99,7 +99,7 @@ void volcanic_cmip_sw2(const ThreadTeam &team, const ConstColumnView &zi,
         half * (tau_w_f(i, ilev_tropp + 1) + ext_ssa_asym * asym_unitless);
   });
 
-  Kokkos::parallel_for(Kokkos::TeamThreadRange(team, ilev_tropp), [&](int kk) {
+  Kokkos::parallel_for(Kokkos::TeamVectorRange(team, ilev_tropp), [&](int kk) {
```
odiazib, line 102: Here, we are using` Kokkos::TeamThreadRange` and `Kokkos::ThreadVectorRange`. We cannot use TeamVectorRange. Is this correct? We only use TeamVectorRange if there are no nested loops.

overfelt: As far as I can tell from the documentation `TeamThreadRange` and `TeamVectorRange` have the same restrictions:
[](https://kokkos.org/kokkos-core-wiki/API/core/policies/TeamThreadRange.html)
[](https://kokkos.org/kokkos-core-wiki/API/core/policies/TeamVectorRange.html)

The descriptions would seem to be the same for each.

If `volcanic_cmip_sw2` is already called in a `parallel_for` to get the team handle, it would seem to me that the third nested `parallel_for` below would be one to many.  But there are no compile warnings or errors.

overfelt: As far as I can tell from the documentation `TeamThreadRange` and `TeamVectorRange` have the same restrictions:
https://kokkos.org/kokkos-core-wiki/API/core/policies/TeamThreadRange.html
https://kokkos.org/kokkos-core-wiki/API/core/policies/TeamVectorRange.html

The descriptions would seem to be the same for each.


If `volcanic_cmip_sw2` is already called in a `parallel_for` to get the team handle, it would seem to me that the third nested `parallel_for` below would be one to many.  But there are no compile warnings or errors.

singhbalwinder: Thanks, @overfelt. For the current architectures, there are only two levels of parallelism, right? Even if we add a third Kokkos parallel_for, it will work but the thirst loop will run in serial. Is that correct?

PR #372, opened 2024-11-05
fix cuda compile error for `aero_model_emissions`

mjs271:
Minor fix to get things compiling on Weaver.

Required making copies of the offending namespace variables that are local to the `aero_model_emissions()` function call, and then passing the local variables to the lower-level functions.


singhbalwinder (approved):
I like your fix! I fixed the same issue differently on PM-GPU.

odiazib (approved):

PR #295, opened 2024-02-09
`mo_sethet` pt 2 - `sethet`

jaelynlitz:
The bulk of the port is there - some resulting questions though:
- what is `wetdep_map`?
- indexes from [`get_spc_ndx` and `get_het_ndx`](https://github.com/eagles-project/e3sm_mam4_refactor/blob/refactor-maint-2.0/components/eam/src/chemistry/mozart/mo_chem_utls.F90) - does that just correlate to [GasId](https://github.com/eagles-project/mam4xx/blob/main/src/mam4xx/aero_modes.hpp#L266)?
- what is necessary in the [last for loop](https://github.com/eagles-project/mam4xx/blob/jaelynlitz/mo_sethet_dev/src/mam4xx/mo_sethet.hpp#L469)? Obviously need to set `het_rates`. Also check for Missing values? Don't need to do the write so don't need `hetratestrg`?
- I see #293 - does `gas_pcnst` == `pcnst` == `gas_wetdep_cnt`?

PR #373, opened 2024-11-08
Final: `mo_sethet`

jaelynlitz:
opened to take over #295 to have a clean starting point to debug from

### DIFF
--- a/src/mam4xx/mo_sethet.hpp
+++ b/src/mam4xx/mo_sethet.hpp
@@ -274,27 +274,21 @@ void sethet(
   // 'H2O2','H2SO4','SO2'.  Options for other species are then removed
   //-----------------------------------------------------------------
 
-  Kokkos::parallel_for(
-        Kokkos::TeamThreadRange(team, pver), [&](int kk) {
-          Kokkos::parallel_for(
-              Kokkos::ThreadVectorRange(team, gas_pcnst),
-              [&](int mm) {
-                het_rates[mm](kk) = 0.0;
-                tmp_hetrates[mm](kk) = 0.0; // initiate temporary array
-              });
+  Kokkos::parallel_for(Kokkos::TeamThreadRange(team, pver), [&](int kk) {
+    Kokkos::parallel_for(
+        Kokkos::ThreadVectorRange(team, gas_pcnst), [&](int mm) {
+          het_rates[mm](kk) = 0.0;
+          tmp_hetrates[mm](kk) = 0.0; // initiate temporary array
         });
-
+  });
 
   for (int mm = 0; mm < gas_wetdep_cnt; mm++) {
     int mm2 = wetdep_map[mm];
-          if (mm2 < 0) {
-            Kokkos::parallel_for(
-                Kokkos::ThreadVectorRange(team, pver),
-                [&](int kk) {
-                  het_rates[mm2](kk) = MISSING;
-                });
-          }
-        }
+    if (mm2 < 0) {
+      Kokkos::parallel_for(Kokkos::ThreadVectorRange(team, pver),
+                           [&](int kk) { het_rates[mm2](kk) = MISSING; });
+    }
+  }
 
   //-----------------------------------------------------------------
   //	... the 2 and .6 multipliers are from a formula by frossling (1938)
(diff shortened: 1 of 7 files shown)


jaelynlitz:
Thanks @odiazib ! Finally ready for a review & merge 😄

src/mam4xx/mo_sethet.hpp
```diff
@@ -11,15 +11,13 @@ namespace mam4 {
 
 namespace mo_sethet {
 
-// const int ktop = ConvProc::Config::ktop;
 constexpr Real avo = haero::Constants::avogadro;
 const Real pi = haero::Constants::pi;
```
odiazib, line 15: `const` to `constexpr`

src/validation/mo_sethet/sethet.cpp
```diff
@@ -0,0 +1,157 @@
...
+void sethet(Ensemble *ensemble) {
+  ensemble->process([=](const Input &input, Output &output) {
+    using View1DHost = typename HostType::view_1d<Real>;
+    using ColumnView = haero::ColumnView;
+    constexpr int pver = mam4::nlev;
+    constexpr int gas_pcnst = mam4::gas_chemistry::gas_pcnst;
+
+    // non-ColumnView input values
+    const Real rlat = input.get_array("rlat")[0]; // need
+    // const Real rlat = -.2320924702;
```
odiazib, line 26: remove it.

odiazib (approved):

src/mam4xx/mo_sethet.hpp
```diff
@@ -176,6 +183,303 @@ void find_ktop(
...
+  }
+
+  //-----------------------------------------------------------------
+  //	... Set rates above tropopause = 0.
+  //-----------------------------------------------------------------
+  team.team_barrier();
+  bool abort = false;
+  for (int mm = 0; mm < gas_wetdep_cnt; mm++) {
+    int mm2 = wetdep_map[mm];
+    Kokkos::parallel_for(Kokkos::TeamVectorRange(team, ktop),
```
jeff-cohere, line 466: There might be a more expedient way to zero out `het_rates[mm2]`.

src/mam4xx/mo_sethet.hpp
```diff
@@ -176,6 +183,303 @@ void find_ktop(
...
+                         [&](int kk) { het_rates[mm2](kk) = 0.0; });
+    Kokkos::parallel_for(Kokkos::TeamVectorRange(team, local_pver),
+                         [&](int kk) {
+                           if (het_rates[mm2](kk) == MISSING)
```
jeff-cohere, line 470: Is this just a sanity check?

jeff-cohere (approved):
I left a couple of comments. Thanks for replacing the other PR with this one, @jaelynlitz .

src/mam4xx/mo_sethet.hpp
overfelt: It looks like `abort` is being set on device and then checked on host. Some values are automatically copied host to device but never back to host.  A parallel_reduction is probably needed instead of a parallel_for.

singhbalwinder (approved):
Thanks for your hard work and getting it to the finish line!

src/mam4xx/mo_sethet.hpp
```diff
@@ -176,6 +183,303 @@ void find_ktop(
...
+          het_rates[mm](kk) = 0.0;
+          tmp_hetrates[mm](kk) = 0.0; // initiate temporary array
+        });
+  });
+
+  for (int mm = 0; mm < gas_wetdep_cnt; mm++) {
+    int mm2 = wetdep_map[mm];
+    if (mm2 < 0) {
+      Kokkos::parallel_for(Kokkos::TeamVectorRange(team, local_pver),
+                           [&](int kk) { het_rates[mm2](kk) = MISSING; });
```
overfelt, line 295: It seems `mm2` is used as an index to `het_rates` if `mm2` is negative.  Maybe it should be `mm` as the index?

overfelt (approved):

src/mam4xx/mo_sethet.hpp
jaelynlitz: Thanks for noticing that! The inequality needs to flip to match the fortran

src/mam4xx/mo_sethet.hpp
jaelynlitz: Gotcha, so I could do a parallel_reduce where if a `MISSING` value is found, add something to the cumulative variable, then on host if the var isn't 0 then abort?

overfelt: Yes, copy the logic that reduces total_rain and change to an int.

PR #376, opened 2024-11-17
Hook up drydep interface.

overfelt:


src/mam4xx/mo_gas_phase_chemdr.hpp
```diff
@@ -82,11 +106,15 @@ void perform_atmospheric_chemistry_and_microphysics(
     const View1D &linoz_PmL_clim_icol, const View1D &linoz_dPmL_dO3_icol,
     const View1D &linoz_dPmL_dT_icol, const View1D &linoz_dPmL_dO3col_icol,
     const View1D &linoz_cariolle_pscs_icol, const Real eccf,
-    const Real adv_mass_kg_per_moles[gas_pcnst],
+    const Real adv_mass_kg_per_moles[gas_pcnst], const Real mmr[gas_pcnst],
+    const Real fraction_landuse[mam4::mo_drydep::n_land_type],
+    const int col_index_season[mam4::mo_drydep::n_land_type],
```
odiazib, line 111: @overfelt For col_index_season, please see PR3108. This PR was merged into the master branch in the scream repo. However, I do not see these changes in the master branch of E3SM. We may need to wait or request for these branches to be synced.

src/mam4xx/seq_drydep.hpp
```diff
@@ -137,6 +137,168 @@ void set_hcoeff_scalar(const Real sfc_temp, Real heff[]) {
...
+     { 100.0, 200.0, 100.0,2000.0,2000.0,2000.0,   0.0,   0.0, 300.0, 150.0, 200.0}, 
+     { 100.0, 150.0, 100.0,1500.0,2000.0,1700.0,   0.0,   0.0, 200.0, 120.0, 140.0},
+     { 100.0,  10.0, 100.0,1000.0,2000.0,1500.0,   0.0,   0.0, 100.0,  50.0, 120.0}, 
+     { 100.0,  10.0,  10.0,1000.0,2000.0,1500.0,   0.0,   0.0,  50.0,  10.0,  50.0}, 
+     { 100.0,  50.0,  80.0,1200.0,2000.0,1500.0,   0.0,   0.0, 200.0,  60.0, 120.0} };
+  for (int i = 0; i < NSeas; ++i) {
+    for (int j = 0; j < NLUse; ++j) {
+        data.rac(i, j) = rac_a[i][j];
+    }
+  }
```
mingxuanwupnnl, line 190: The way of initialization works on CPU but probably not for GPU. Need to fix it. @odiazib @singhbalwinder and I discussed it before, I @odiazib has ways to fix it.

src/mam4xx/mo_gas_phase_chemdr.hpp
```diff
@@ -273,8 +301,26 @@ void perform_atmospheric_chemistry_and_microphysics(
...
+        pressure_10m,     // 10-meter pressure [Pa]
+        spec_hum,         // specific humidity [kg/kg]
+        wind_speed,       // 10-meter wind spped [m/s]
+        rain,             // rain content [??]
+        snow,             // snow height [m]
+        solar_flux,       // direct shortwave surface radiation [W/m^2]
+        mmr,              // constituent MMRs [kg/kg]
+        dvel,             // deposition velocity [1/cm/s]
+        dflx              // deposition flux [1/cm^2/s]
+    );
```
mingxuanwupnnl, line 323: need to add `if (kk == nlev)` for drydep_xactive. Only gas dry dep fluxes and velocity are only calculated at bottom layer. 

Some input vars need calculation. For example, wind_speed, it is calculated as sqrt(ubot^2 + vbot^2).

mingxuanwupnnl (changes requested):
Thank you @overfelt. I just left some comments (may edit them later).

singhbalwinder:
I should have tested it before pushing. I will test it and fix the build errors.

src/mam4xx/mo_sethet.hpp
```diff
@@ -481,6 +481,10 @@ void sethet_detail(
     int mm2 = wetdep_map[mm];
     Kokkos::parallel_for(Kokkos::TeamVectorRange(team, ktop),
                          [&](int kk) { het_rates(kk, mm2) = 0.0; });
+  }
+  Kokkos::fence();
+  for (int mm = 0; mm < gas_wetdep_cnt; mm++) {
+    int mm2 = wetdep_map[mm];
```
singhbalwinder, line 487: @overfelt @odiazib @mjs271 : I was getting an error (`sethet: het_rates (wet dep) not set for het reaction number`)   [see line 498 of this file] and I added this fence. Please let me know if there is a better way to fix it. I believe the error is due to a race condition where `abort` becomes > 0.

odiazib: We should use `team.team_barrier()`, as this part of the code is within a parallel_for.

overfelt: Since you are modifying the code anyway, doing a ThreadVectorRange inside a TeamThreadRange as on line 298 is the way to go. And, yes, team_barrier() is the right call in this case. 

A team_barrier() right before the abort check on line 496 might be overkill, but there are already five team_barrier()s in this function, maybe one more would not hurt.

src/mam4xx/mo_gas_phase_chemdr.hpp
```diff
@@ -331,8 +358,26 @@ void perform_atmospheric_chemistry_and_microphysics(
...
+        pressure_10m,     // 10-meter pressure [Pa]
+        spec_hum,         // specific humidity [kg/kg]
+        wind_speed,       // 10-meter wind spped [m/s]
+        rain,             // rain content [??]
+        snow,             // snow height [m]
+        solar_flux,       // direct shortwave surface radiation [W/m^2]
+        qq,               // constituent MMRs [kg/kg]
+        dvel,             // deposition velocity [1/cm/s]
+        dflx              // deposition flux [1/cm^2/s]
+    );
```
mingxuanwupnnl, line 380: gas dry dep is only executed for bottom layer. I think we need to add if condition here. Some of the input vars for dry dep needs calculation. Currently it is done in EAMxx branch. We might consider move them to MAMxx code if it is the preferred practice.

mingxuanwupnnl (changes requested):

PR #377, opened 2024-11-19
Fix else branch indexing in dust_emis()

mjs271:
Corrected to match `if` branch and ensure the dust indices of `cflux` are properly zeroed-out in the case that `soil_erodibility < soil_erod_threshold`


singhbalwinder (approved):
Thanks for fixing this!

PR #379, opened 2024-11-19
MAM4xx: make a local copy of nlev to avoid compilation error in CUDA.

odiazib:


singhbalwinder (approved):
Thanks for the fix!

jeff-cohere (approved):
Looks great!

PR #383, opened 2024-12-10
H2O2 washout rates

jaelynlitz:


jaelynlitz:
In the first iteration of `sethet`, we were calculating an additional level that the Fortran was not (see [note in yaml](https://github.com/eagles-project/mam_x_validation/blob/90edf8a8f07927362e508f6fdcff882292e06ffb/mo_sethet/mam_sethet_ts_355.py#L4) due to our calculation of `ktop`. This PR artificially uses `ktop+1` when we're setting the rates "above tropopause". Is this the path forward we want to use? This PR passes with all the additional validation data, unclear if this resolves the original issue they were seeing in the simulations.

mjs271 (approved):
Looks great--thank you!

overfelt (approved):

jeff-cohere (approved):
I don't know the consequences of using this extra level, but I'm all for trying it out! Thanks, @jaelynlitz !

jaelynlitz:
CI failed on Deception, but worked by hand with the new modules... I'm inclined to hit merge since the failure should resolve once I update the modules in CI

PR #384, opened 2024-12-16
Set to valid data to avoid divide-by-zero in unit test.

overfelt:


jaelynlitz (approved):
PNNL test failures are unrelated - LGTM

PR #385, opened 2024-12-16
Date from yaml file is MMDDYY format. Extract month for test.

overfelt:


jaelynlitz (approved):
PNNL test failures are unrelated - LGTM

PR #386, opened 2024-12-16
Loop index caused solsym array to be indexed past end-of-array.

overfelt:


src/mam4xx/mo_chm_diags.hpp
```diff
@@ -297,7 +297,7 @@ void chm_diags(
   // accumulators
   // if (history_aerosol.and..not .history_verbose) then
 
-  for (int nn = lchnk; nn < pcnst; nn++) {
+  for (int nn = lchnk; nn < gas_pcnst; nn++) {
```
jaelynlitz, line 300: Maybe the real discrepancy is that in this file we set `pcnst` to 80, but in the validation test, we use `pcsnt` = 41? I agree that `solsym` is length `gas_pcnst` but `fldcw` can technically be length `pcnst` ?

overfelt: Yes, there is something wrong with the length of fldcw. There was a note that setting the length to 80 was somewhat of a guess as it was not obvious in the Fortran code what length to use.  Maybe we should ask Balwinder when he gets back?

jaelynlitz: That sounds like a good plan

PR #387, opened 2024-12-16
Add limiting to avoid indexing past end of array.

overfelt:


jaelynlitz (approved):
Good catch! LGTM, PNNL test failures are the same unrelated ones

PR #394, opened 2025-01-13
fix work array size for wet_dep() and treat isprx as int ptr

mjschmidt271:
This appears to fix the out-of-bounds issue that's causing the E3SM standalone wet deposition test to fail.


singhbalwinder (approved):
Thanks, Mike!

jeff-cohere (approved):

PR #396, opened 2025-01-15
Fixes for a deterministic world. 

overfelt:


jeff-cohere (approved):
Looks good to me. The dynamic you've described with certain cases involving `Kokkos::single` is confusing, but if you're getting good results, these changes seems worth trying. The switch to `TeamVectorRange` from `ThreadVectorRange` makes perfect sense.

Issue #393, opened 2025-01-10
Failing tests after module upgrade


jaelynlitz:
- [ ] `kohler_unit_test` is failing in CI in the double Release build and I cannot reproduce by hand on Deception.

```
 16/629 Test  #16: kohler_unit_tests ..................................................................***Failed    1.96 sec
Calling initialize_kokkos
 ExecSpace name: Cuda
 ExecSpace initialized: yes
 active avx set: 
 compiler id: GCC
 FPE support is enabled, current FPE mask: 0 (NONE)
 #host threads: 1

[2024-12-27 12:32:29.458] [kohler functions] [info] Accounting for temperature changes causes <= 0.1594234332126723 relative difference in surface tension.
[2024-12-27 12:32:29.458] [kohler functions] [info] Accounting for temperature changes causes <= 0.34055359583775985 relative difference in Kelvin droplet coefficient.
reading true solutions from data file: /people/svceagles/gitlab/7814/haero_Release_double/src/tests/data/mm_kohler_roots.txt

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
kohler_unit_tests is a Catch v2.13.8 host application.
Run with -? for options

-------------------------------------------------------------------------------
kohler_verificiation
  polynomial_properties
-------------------------------------------------------------------------------
/people/svceagles/gitlab/7814/haero_Release_double/src/tests/kohler_unit_tests.cpp:133
...............................................................................

/people/svceagles/gitlab/7814/haero_Release_double/src/tests/kohler_unit_tests.cpp:133: FAILED:
due to a fatal error condition:
  SIGFPE - Floating point error signal

===============================================================================
test cases: 2 | 1 passed | 1 failed
assertions: 3 | 2 passed | 1 failed

EKAT is not managing resources.
RUN: OMP_PROC_BIND=spread OMP_PLACES=threads ./kohler_unit_tests --use-colour no
FROM: /qfs/people/svceagles/gitlab/7814/haero_Release_double/build/src/tests
```

- [X] `validate_aer_rad_props_sw_ts_355` also fails intermittently, likely a race condition somewhere.
```
L1 rel_error 5.787420836758631e-07
L2 rel_error 5.786842910970662e-07
Linf rel_error 5.786842910820387e-07
final pass array = [ True  True False  True False]
Traceback (most recent call last):
[...]

PR #398, opened 2025-01-22
Draft: Debug kohler unit test

jaelynlitz:
#393


jaelynlitz:
We are de-prioritizing this - will return if this is reproduced on SNL machines

PR #414, opened 2025-03-17
Oscar/ndrop top lev

odiazib:
`top_lev` was hard-coded in ndrop and used Fortran indexing. I have made it an input, so we can use the value from the namelist in the emaxx interface. In the emaxx interface, we are using C++ indexing.


singhbalwinder (approved):
Thanks , Oscar for taking care of it so quickly!

PR #419, opened 2025-04-09
Using views for scavimptblvol and scavimptblnum.

odiazib:
Using views instead of real arrays for scavimptblnum and scavimptblvol. In addition, I deleted a block of code that is not used by eamxx.


singhbalwinder (approved):
Thanks for cleaning up this code!

PR #420, opened 2025-04-14
Frontier fix. 

odiazib:
Use real arrays for dvel and dflx, and move the initialization of photo to the Emaxx interface.


singhbalwinder (approved):
Thanks for taking care of it so quickly!

PR #439, opened 2025-05-06
Removing .data() from calcsize routines.

odiazib:
I am removing `view.data()` from the `calcisize` routines. I am using a template parameter `VectorType` to replace` Real* var ` because we are using Real arrays a few times.

I compiled and ran the main branch of mam4xx in Frontier, and four validation tests failed:

```
    16 - validate_stand_modal_aero_calcsize_sub_update_ptend (Failed)
    18 - validate_stand_calcsize_aero_model_wetdep_ts_379 (Failed)
    492 - validate_chm_diags_ts_355 (Failed)
    524 - validate_gas_washout_merged (Failed)
```

I note that this PR fixes `validate_stand_modal_aero_calcsize_sub_update_ptend` and `validate_stand_calcsize_aero_model_wetdep_ts_379`. Thus, using raw pointers is producing issues in the AMD backend. I will continue replacing .data() in a follow-up PR.

See [PR 7335](https://github.com/E3SM-Project/E3SM/pull/7335) for eamxx testing.

### DIFF
--- a/src/validation/aero_model/aero_model_calcsize_water_uptake_dr.cpp
+++ b/src/validation/aero_model/aero_model_calcsize_water_uptake_dr.cpp
@@ -92,8 +92,8 @@ void aero_model_calcsize_water_uptake_dr(Ensemble *ensemble) {
                 const auto dqqcwdt_k =
                     Kokkos::subview(dqqcwdt, kk, Kokkos::ALL());
                 modal_aero_calcsize::modal_aero_calcsize_sub(
-                    state_q_k.data(), // in
-                    qqcw_k.data(),    // in/out
+                    state_q_k, // in
+                    qqcw_k,    // in/out
                     dt, cal_data,
                     // outputs
                     dgncur_i.data(), dgncur_c, ptend_q_k.data(),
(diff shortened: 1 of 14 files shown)


src/validation/aero_model/aero_model_calcsize_water_uptake_dr.cpp
```diff
@@ -110,10 +108,9 @@ void aero_model_calcsize_water_uptake_dr(Ensemble *ensemble) {
 
                 mam4::water_uptake::modal_aero_water_uptake_dr(
                     cal_data.nspec_amode, cal_data.specdens_amode,
-                    cal_data.spechygro, cal_data.lspectype_amode,
-                    state_q_k.data(), temperature(kk), pmid(kk), cldn(kk),
-                    dgncur_i.data(), dgnumwet_kk.data(), qaerwat_kk.data(),
-                    wetdens_kk.data());
+                    cal_data.spechygro, cal_data.lspectype_amode, state_q_k,
+                    temperature(kk), pmid(kk), cldn(kk), dgncur_i.data(),
```
mjschmidt271, line 112: it looks like these may be leftovers?

mjschmidt271 (changes requested):
nice work! just have 1 comment for you

src/validation/aero_model/aero_model_calcsize_water_uptake_dr.cpp
singhbalwinder: Let's do smaller PRs, as it makes it easy to track if something goes wrong.

singhbalwinder (approved):
Thanks, Oscar!

src/validation/aero_model/aero_model_calcsize_water_uptake_dr.cpp
mjschmidt271: No need to hold up this PR. I figured it was likely fine but wanted to point it out 🙂

mjschmidt271 (approved):

PR #440, opened 2025-05-07
Add files via upload

polunma:


meng630 (approved):

mjschmidt271 (changes requested):
This seems inaccurate, since this is a code repository that is still being developed and is not an "account of work [...]".

Is this sentiment not covered by the license?

If it turns out this needs to be included, please add to the bottom of the README, as in the [haero repo](https://github.com/eagles-project/haero/blob/main/README.md)

PR #446, opened 2025-05-16
Adding diagnostics for so4 aqueous chemistry and h2so4 uptake.

overfelt:
And there were a few compile warnings about unused variables.


singhbalwinder (approved):
Looks great! Thanks James for fixing other bugs/warnings along the way.

PR #447, opened 2025-05-16
Adds logic to avoid very small numbers for dryvol

singhbalwinder:
For very high-resolution runs, we occasionally encounter small negative values for aerosol dry volume in some grid cells. Although these values are close to zero, they can lead to extremely large results during mathematical operations like division or exponentiation. To prevent this, we introduced a condition that detects small negative dry volumes and skips the subsequent calculations. In such cases, we set all optical parameters to zero.

Another `if` condition is fixed, and an `abs` is added to mimic the original MAM4 Fortran code.

This fix enabled stable long-term CIME simulations at high resolutions, such as ne120 and ne256.

[NBFB]


odiazib (approved):

overfelt (approved):

kaizhangpnl (approved):
The fix enables the ne120pg2 model to run stably on Chrysalis for more than three (model) months.

PR #452, opened 2025-06-04
Expose non-vertical integrated values.

overfelt:


src/mam4xx/mo_gas_phase_chemdr.hpp
```diff
@@ -436,8 +437,27 @@ void perform_atmospheric_chemistry_and_microphysics(
           conversions::mmr_from_vmr(vmr_so4, adv_mass_kg_per_moles[ll]);
       aqh2so4_flx[m] =
           conversions::mmr_from_vmr(vmr_h2s, adv_mass_kg_per_moles[ll]);
+      if (aqso4_lev.size()) {
+        Kokkos::parallel_for(
+            Kokkos::TeamVectorRange(team, nlev), [&](const int kk) {
+              const Real aqso4 = dqdt_aqso4(ll, kk);
+              const Real vmr_so4 = aero_model::calc_sfc_flux(aqso4, pdel(kk));
```
singhbalwinder, line 444: I am not sure if we need to call `calc_sfc_flux` here. Conversion from vmr to mmr should be enough for this quantity. @kaizhangpnl might know more about how he would like to use this in his analysis or what makes more sense from an evaluation perspective.

kaizhangpnl (approved):

src/mam4xx/mo_gas_phase_chemdr.hpp
kaizhangpnl: > I am not sure if we need to call `calc_sfc_flux` here. Conversion from vmr to mmr should be enough for this quantity. @kaizhangpnl might know more about how he would like to use this in his analysis or what makes more sense from an evaluation perspective.

For 3D variables, we often directly look at the tendency (kg/kg/s), so the vmr to mmr is enough (I find it already revised).

singhbalwinder (approved):
Thanks for quickly taking care of this request!

PR #454, opened 2025-06-10
Frontier Fix

odiazib:
Using views instead of arrays fixed the Failed to allocate file: Bad file descriptor error in Frontier.


singhbalwinder (approved):
Awesome!Thanks for taking care of it!

PR #455, opened 2025-06-19
Changes to expose diagnostic quantities to EAMxx (`*_sfgaex1, *_sfgaex2, *_sfnnuc, *_sfcoag`)

mjschmidt271:
Nothing too exciting here--mainly plumbing.

I changed the arrays into views, since it seemed a bit tidier to do so.

I'll admit I'm not fully up to date on our best practices for passing variables around, so please let me know about any of this that might be inefficient or ugly.

### DIFF
--- a/src/mam4xx/mo_gas_phase_chemdr.hpp
+++ b/src/mam4xx/mo_gas_phase_chemdr.hpp
@@ -153,8 +153,7 @@ void perform_atmospheric_chemistry_and_microphysics(
     Real dvel[gas_pcnst], // deposition velocity [cm/s]
     Real dflx[gas_pcnst],
     // out
-    const View3D &qgcm_tendaa,
-    const View3D &qqcwgcm_tendaa,
+    const View3D &qgcm_tendaa, const View3D &qqcwgcm_tendaa,
     mam4::Prognostics &progs) {
 
   const int nlev = mam4::nlev;
--- a/src/mam4xx/mo_gas_phase_chemdr.hpp
+++ b/src/mam4xx/mo_gas_phase_chemdr.hpp
@@ -28,6 +28,7 @@ using View2D = DeviceType::view_2d<Real>;
 using ConstView2D = DeviceType::view_2d<const Real>;
 using View1D = DeviceType::view_1d<Real>;
 using ConstView1D = DeviceType::view_1d<const Real>;
+using View3D = DeviceType::view_3d<Real>;
 
 KOKKOS_INLINE_FUNCTION
 void mmr2vmr_col(const ThreadTeam &team, const haero::Atmosphere &atm,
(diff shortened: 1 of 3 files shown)


src/mam4xx/mo_gas_phase_chemdr.hpp
```diff
@@ -148,7 +149,9 @@ void perform_atmospheric_chemistry_and_microphysics(
     const View1D &work_set_het, const seq_drydep::Data &drydep_data,
     const View1D &aqso4_flx, const View1D &aqh2so4_flx,
     Real dvel[gas_pcnst], // deposition velocity [cm/s]
-    Real dflx[gas_pcnst], mam4::Prognostics &progs) {
+    Real dflx[gas_pcnst],
+    // out
+    View3D &qgcm_tendaa, View3D &qqcwgcm_tendaa, mam4::Prognostics &progs) {
```
jeff-cohere, line 154: I think passing these Views this way is the right way to do it in spirit. I don't know what the "blessed" way of doing it these days is, because
1. I believe you can omit the `&`, as Views wrap pointers
2. IIRC, you can also stick a `const` on them even when you change their contents, because their `const`ness is not determined by that of their underlying storage.

But I don't see anything wrong with this.

jeff-cohere (approved):

singhbalwinder:
Thanks, Mike. I also think we should use views rather than arrays, but at the same time, we should proceed with caution since we are running into a lot of issues with AMD vs. NVIDIA GPUs. @odiazib is working on resolving these issues.

I think we should make minimal changes just to get these diags and do the rest of the improvements in a PR in the future.

For these diagnostics, we can create two (ncol,pver,pcnst) dimensioned variables in EAMxx. In MAM4xx code, they will just be dimensioned `gas_pcnst`. Here is what I think we should do:
1. Create two variables (to store `sfgaex1` and `sfgaex2` diags) with dimensions `gas_pcnst` in `mam4_amicphys.hpp` to expose `qgcm_tendaa` for indices 0 and 1. That is `var1 = qgcm_tendaa[:][0] * pdel/gravit` and `var12= qgcm_tendaa[:][1] pdel/gravit`
2. Bring these 2 variables to the interface level in EAMxx.
3. In EAMxx, we can use "get_field_out" to output these variables by storing them in variables dimensioned (ncol,pver,pcnst)

We need to decide whether to create a new dimension `gas_pcnst` or just use `pcnst`. `pcnst` already exists and it is 41. `gas_pcnst` may not exist yet (it is 31). We can decide it later.

Let me know if it doesn't make sense, as I just eyeballed the code and may have missed some details.

singhbalwinder:
We can also capture `_sfnnuc1` and `_sfcoag1` by declaring two more variables in this PR:
```
pdel_fac = pdel/gravit
var1 = qgcm_tendaa[:][0] * pdel_fac; //sfgaex1
var2 = qgcm_tendaa[:][1] * pdel_fac; //sfgaex1
var3 = qgcm_tendaa[:][2] * pdel_fac; //sfnnuc1
var4 = qgcm_tendaa[:][3] * pdel_fac; //sfcoag1
```
We can choose some reasonable names for var1, var2, var3, and var4.

PR #457, opened 2025-06-23
Implementation of AQ and GS diagnostics.

overfelt:


src/mam4xx/gas_phase_chemistry.hpp
```diff
@@ -41,7 +41,7 @@ void gas_phase_chemistry(
     const Real invariants[nfs], const int (&clsmap_4)[gas_pcnst],
     const int (&permute_4)[gas_pcnst], const Real het_rates[gas_pcnst],
     // out
-    Real (&qq)[gas_pcnst], Real (&vmr0)[gas_pcnst]) {
+    Real (&qq)[gas_pcnst]) {
```
singhbalwinder, line 44: Thanks for changing the code to keep vmr0 in just one file.

singhbalwinder (approved):
Looks good! Thanks!

src/mam4xx/gas_phase_chemistry.hpp
singhbalwinder: @overfelt : The vmr0 change is causing BFB tests to fail. I just tested it, and the tests pass when I change vmr0 treatment back to the original code.

singhbalwinder:
Your latest commit fixes it!

src/mam4xx/mo_gas_phase_chemdr.hpp
```diff
@@ -329,6 +343,10 @@ void perform_atmospheric_chemistry_and_microphysics(
     // the logic below is taken from the aero_model_gasaerexch
     // subroutine in eam/src/chemistry/modal_aero/aero_model.F90
 
+    // Store mixing ratios before aqueous chemistry changes the mixing ratios
+    Real vmr1[gas_pcnst] = {};
+    for (int i = 0; i < gas_pcnst; ++i)
+      vmr1[i] = vmr[i];
```
singhbalwinder, line 349: Can we change the variable name from `vmr1` to `vmr_bef_aq_chem`, since this is a new variable?

PR #458, opened 2025-06-24
Expose the activation fractions to make them settable from E3SM.

overfelt:
Yet another change to a function signature in order to be able to set these values form the E3SM input.


singhbalwinder (approved):
Looks good. Thanks, @overfelt !

PR #459, opened 2025-06-24
Make output so4 and h2 so4 aqueous chemistry column integrated flux diagnostic optional.

overfelt:


src/mam4xx/mo_chm_diags.hpp
```diff
@@ -314,7 +314,7 @@ void chm_diags(
   // accumulators
   // if (history_aerosol.and..not .history_verbose) then
 
-  for (int nn = lchnk; nn < pcnst; nn++) {
+  for (int nn = lchnk; nn < gas_pcnst; nn++) {
```
singhbalwinder, line 317: I don't quite understand this change. Can we verify if it is okay to change it to gas_pcnst?
If it takes time, let's just revert this change and investigate it later.

src/validation/modal_aero_amicphys_subareas/form_gcm_of_gases_and_aerosols_from_subareas.cpp
```diff
@@ -72,7 +72,8 @@ void form_gcm_of_gases_and_aerosols_from_subareas(Ensemble *ensemble) {
       std::cout << "n = " << n << "\n";
       afracsub[j] = afracsub_[j - 1];
       std::cout << "afracsub[j] = " << afracsub[j] << "\n";
-      std::cout << "afracsub_[n] = " << afracsub_[n] << "\n";
+      if (j < nsubarea)
+        std::cout << "afracsub_[j] = " << afracsub_[j] << "\n";
```
singhbalwinder, line 76: Could you please explain this change? It looks like a bug fix to me.

singhbalwinder (changes requested):
Thanks, James! I left some minor comments/questions.

src/mam4xx/convproc.hpp
```diff
@@ -567,8 +567,8 @@ class ConvProc {
     // Set nucleation-specific config parameters.
     config_ = convproc_config;
     Kokkos::resize(scratch1Dviews[q], config_.nlev * aero_model::pcnst);
-    Kokkos::resize(scratch1Dviews[mu], config_.nlev);
-    Kokkos::resize(scratch1Dviews[md], config_.nlev);
+    Kokkos::resize(scratch1Dviews[mu], config_.nlev + 1);
+    Kokkos::resize(scratch1Dviews[md], config_.nlev + 1);
```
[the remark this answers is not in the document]
overfelt: This was causing my address sanitizer tests to fail. It is not directly related to the diagnostic output.

src/mam4xx/mo_chm_diags.hpp
singhbalwinder: Let's create a GitHub issue for this and investigate. We can revert this change for now. We will fix it after we understand the right way to do it.

PR #461, opened 2025-06-26
Add a script for automated model performance comparison between branches

meng630:


jeff-cohere (approved):
This looks good to me. You have been using these scripts to do performance analysis, right? If so, let's merge them in.

PR #463, opened 2025-07-07
Pass gas-aero exchange tendencies to EAMxx to output as diagnostics

mjschmidt271:
This is a re-do on #455 

This follows the pattern implemented by @overfelt  of using the `MicrophysDiagnosticArrays` struct to hold the diagnostics for output.

### DIFF
--- a/src/mam4xx/mam4_amicphys.hpp
+++ b/src/mam4xx/mam4_amicphys.hpp
@@ -3,7 +3,6 @@
 
 #include <mam4xx/aging.hpp>
 #include <mam4xx/coagulation.hpp>
-#include <mam4xx/diagnostic_arrays.hpp>
 #include <mam4xx/gas_chem_mechanism.hpp>
 #include <mam4xx/gasaerexch.hpp>
 #include <mam4xx/nucleation.hpp>
@@ -13,7 +12,7 @@ namespace mam4 {
 
 namespace microphysics {
 
-using View1D = DeviceType::view_1d<Real>;
+using View2D = DeviceType::view_2d<Real>;
 
 // number of constituents in gas chemistry "work arrays"
 using mam4::gas_chemistry::gas_pcnst;


src/mam4xx/mam4_amicphys.hpp
```diff
@@ -2119,7 +2120,13 @@ void modal_aero_amicphys_intr(
...
     const Real qv, const Real cld,
     // in/out
-    Real qq[gas_pcnst], Real qqcw[gas_pcnst],
+    Real (&qq)[gas_pcnst], Real (&qqcw)[gas_pcnst],
+    // Diagnostics (out)
+    Real (&gas_aero_exchange_condensation)[gas_pcnst],
+    Real (&gas_aero_exchange_renaming)[gas_pcnst],
+    Real (&gas_aero_exchange_nucleation)[gas_pcnst],
+    Real (&gas_aero_exchange_coagulation)[gas_pcnst],
+    Real (&gas_aero_exchange_renaming_cw)[gas_pcnst],
```
singhbalwinder, line 2129: We can remove `gas_aero_exchange_renaming_cw` for now until the eval team requests this field.

src/mam4xx/mo_gas_phase_chemdr.hpp
```diff
@@ -388,14 +388,54 @@ void perform_atmospheric_chemistry_and_microphysics(
...
+        diag_arrays.gas_aero_exchange_coagulation(i, kk) =
+            gas_aero_exchange_coagulation[i];
+      }
+    }
+    if (diag_arrays.gas_aero_exchange_renaming_cw.size()) {
+      for (int i = 0; i < gas_pcnst; ++i) {
+        diag_arrays.gas_aero_exchange_renaming_cw(i, kk) =
+            gas_aero_exchange_renaming_cw[i];
+      }
+    }
```
singhbalwinder, line 437: I am not sure if it makes sense to convert this for-loop to a parallel_for or not. Also, we can create a function for the parallel_for as it is repeated multiple times. Something like:
```
if (diag_arrays.gas_aero_exchange_coagulation.size()) 
   populate_diags(gas_aero_exchange_coagulation, diag_arrays.gas_aero_exchange_coagulation)
```
We can do this in a follow-up PR as well.

singhbalwinder (changes requested):
Thanks, Mike for taking care of it so quickly. I have left some minor comments/suggestions.

singhbalwinder (approved):

src/mam4xx/mo_gas_phase_chemdr.hpp
singhbalwinder: We can consider it in a follow-up PR.

Once you are ready, let me know, and I can run this on Frontier before we start merging this.

src/mam4xx/mam4_amicphys.hpp
```diff
@@ -2119,7 +2120,13 @@ void modal_aero_amicphys_intr(
     const Real pmid, const Real pdel, const Real zm, const Real pblh,
     const Real qv, const Real cld,
     // in/out
-    Real qq[gas_pcnst], Real qqcw[gas_pcnst],
+    Real (&qq)[gas_pcnst], Real (&qqcw)[gas_pcnst],
+    // Diagnostics (out)
+    Real (&gas_aero_exchange_condensation)[gas_pcnst],
```
odiazib, line 2125: I wonder if we should use views instead of real arrays.

src/mam4xx/mo_gas_phase_chemdr.hpp
```diff
@@ -388,14 +388,54 @@ void perform_atmospheric_chemistry_and_microphysics(
     }
     // do aerosol microphysics (gas-aerosol exchange, nucleation,
     // coagulation)
+    Real gas_aero_exchange_condensation[gas_pcnst] = {};
```
odiazib, line 391: Can we use views directly instead of creating temporary arrays?

src/mam4xx/mo_gas_phase_chemdr.hpp
```diff
@@ -388,14 +388,54 @@ void perform_atmospheric_chemistry_and_microphysics(
...
+        // diagnostics (out)
+        gas_aero_exchange_condensation, gas_aero_exchange_renaming,
+        gas_aero_exchange_nucleation, gas_aero_exchange_coagulation,
+        gas_aero_exchange_renaming_cw,
         // in
         vmr0, vmr_pregas, vmr_precld, dgncur_a_kk, dgncur_awet_kk, wetdens_kk);
 
+    if (diag_arrays.gas_aero_exchange_condensation.size()) {
+      for (int i = 0; i < gas_pcnst; ++i) {
+        diag_arrays.gas_aero_exchange_condensation(i, kk) =
```
odiazib, line 410: Can we use views directly instead of creating temporary arrays?

src/mam4xx/mam4_amicphys.hpp
singhbalwinder: We can use views since this is new code. I am hesitant to change too much due to the machine issues we are encountering. I will start another PR, right after this one goes in to change these to views. That way, it will be a two-step process, but easier to debug if we get into any issues.

src/mam4xx/mam4_amicphys.hpp
```diff
@@ -2347,16 +2349,20 @@ void modal_aero_amicphys_intr(
...
-    gas_aero_exchange_renaming[icnst] = qgcm_tendaa[icnst][1];   // renaming
-    gas_aero_exchange_nucleation[icnst] = qgcm_tendaa[icnst][2]; // nucleation
-    gas_aero_exchange_coagulation[icnst] = qgcm_tendaa[icnst][3]; // coagulation
-    gas_aero_exchange_renaming_cloud_borne[icnst] =
-        qqcwgcm_tendaa[icnst][0]; // renaming
-  }
+  // Lambda to copy tendencies into a 1D view of all gas and aerosol species, if allocated
+  auto assign_if_allocated = [&]( const auto& view, const auto& tend, const int idx, const int extent) {
+    if (view.data() != nullptr) {
+      for (int i = 0; i < extent; ++i) {
```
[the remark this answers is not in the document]
odiazib: However, we need to pass team as input.

src/mam4xx/mo_gas_phase_chemdr.hpp
```diff
@@ -386,13 +386,32 @@ void perform_atmospheric_chemistry_and_microphysics(
...
+
+    //Lambda function to safely create subviews of diagnostics arrays
+    auto safe_subview = [&](const auto& view2d) {
+      return (view2d.data() != nullptr)
+        ? ekat::subview(view2d, kk)
+        : decltype(ekat::subview(view2d, kk))();
+    };
+
+    // Create subviews of diagnostics arrays for the current vertical level only if allocated
+    auto gas_aero_exchange_condensation = safe_subview(diag_arrays.gas_aero_exchange_condensation);
```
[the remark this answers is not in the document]
singhbalwinder: I attempted to capture this logic in the lambda function above. I used lambda to avoid code duplication as much as I could. We can change `auto` to `view_1d` if we like to be more explicit.

src/mam4xx/mam4_amicphys.hpp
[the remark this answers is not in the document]
singhbalwinder: I will change it to a parallel_for. Neither PM nor Frontier is working for me at this point. Once I have access, I will make this change. Thanks.

PR #464, opened 2025-07-10
A few minor housekeeping additions to CI Workflows

mjschmidt271:
The most notable change is marking the `clang-format` check `continue-on-error` so that the other checks still provide useful info when the format check fails.

Also changed the name of the cpu test to conform to the format being used and added `ext/` and top-level CMake files to the list that triggers testing on a PR.

Finally, I messed up the formatting on `drydep.hpp` to hopefully trigger the behavior I'm looking for


singhbalwinder (approved):
Thanks, Mike! I like the idea of running the rest of the tests even if format check fails.

PR #465, opened 2025-07-16
Updates for PR 7501 in E3SM

odiazib:
See [PR 7501](https://github.com/E3SM-Project/E3SM/pull/7501/)

### DIFF
--- a/src/mam4xx/vertical_interpolation.hpp
+++ b/src/mam4xx/vertical_interpolation.hpp
@@ -15,7 +15,8 @@ using View1DInt = DeviceType::view_1d<int>;
 KOKKOS_INLINE_FUNCTION
 void vert_interp(const ThreadTeam &team, int levsiz, int pver,
                  const ConstView1D &pin, const ConstView1D &pmid,
-                 const View1D &datain, const View1D &dataout) {
+                 const View1D &datain, const View1D &dataout,
+                 const int unit_factor_pin=1) {
   const int zero = 0;
 
   Kokkos::parallel_for(Kokkos::TeamVectorRange(team, pver), [&](int k) {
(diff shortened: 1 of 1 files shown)


src/mam4xx/vertical_interpolation.hpp
```diff
@@ -14,8 +14,9 @@ using View1DInt = DeviceType::view_1d<int>;
 
 KOKKOS_INLINE_FUNCTION
 void vert_interp(const ThreadTeam &team, int levsiz, int pver,
-                 const View1D &pin, const ConstView1D &pmid,
-                 const View1D &datain, const View1D &dataout) {
+                 const ConstView1D &pin, const ConstView1D &pmid,
+                 const View1D &datain, const View1D &dataout,
+                 const int unit_factor_pin=1) {
```
singhbalwinder, line 19: I like constViews wherever we can use them as they make the code easier to read. Can `datain` also be a `ConstView1D`? `levsiz` and `pver` can be `const`.

src/mam4xx/vertical_interpolation.hpp
```diff
@@ -25,21 +26,21 @@ void vert_interp(const ThreadTeam &team, int levsiz, int pver,
     int kkstart = zero;
     // // Store level indices for interpolation
     for (int kk = kkstart; kk < levsiz - 1; ++kk) {
-      if (pin(kk) < pmid(k) && pmid(k) <= pin(kk + 1)) {
+      if (pin(kk)*unit_factor_pin < pmid(k) && pmid(k) <= pin(kk + 1)*unit_factor_pin) {
```
singhbalwinder, line 29: `kkstart` can be `constexpr` or `const`

src/mam4xx/vertical_interpolation.hpp
```diff
@@ -25,21 +26,21 @@ void vert_interp(const ThreadTeam &team, int levsiz, int pver,
...
     }   // end for
 
-    if (pmid(k) < pin(0)) {
+    if (pmid(k) < pin(0)*unit_factor_pin) {
       EKAT_KERNEL_ASSERT_MSG(
           pin(0) != 0.0,
           "Error: Division by zero. The value of pin(0) is zero.\n");
-      dataout(k) = datain(0) * pmid(k) / pin(0);
-    } else if (pmid(k) > pin(levsiz - 1)) {
+      dataout(k) = datain(0) * pmid(k) / pin(0)/unit_factor_pin;
```
singhbalwinder, line 38: I didn't pay attention to the logic behind `unit_factor_pin`, but it is always multiplied to `pin`. Here it is divided. Is this correct? If it is always multiplied, we can do it in one place at the beginning of this function.

odiazib: line 38 is equivalent to
`dataout(k) = datain(0) * pmid(k) / (pin(0)*unit_factor_pin);`

PR #467, opened 2025-07-23
Remove wrong team barrier in perform atm chem/micro

bartgol:
Barriers cannot be inside a nested pfor loop.

Removing this line resolves a hang I am observing when running an eamxx test with the new EKAT. I have no idea why this is not causing a hang in master.


singhbalwinder (approved):
Thanks for finding and fixing it so quickly!

PR #470, opened 2025-07-23
MAM4xx: Adding nested loops back.

odiazib:


overfelt (approved):

singhbalwinder (approved):
Thanks for fixing it, Oscar! Thanks for noticing it, James! It was a good catch!

PR #476, opened 2025-08-20
TESTING - AMD/HIP Autotesting

mjschmidt271:
DO NOT MERGE


mjschmidt271:
To update: I'm debugging a little weirdness with the AMD runners, but this is otherwise good to go

PR #479, opened 2025-10-07
Refactor the code to eliminate a couple of temporary arrays. 

overfelt:


jeff-cohere:
Since this code affects the ACI and seems like an improvement, I'm interested in getting it merged. Now that the Autotester is working again, how does one rerun it on a PR like this one?

PR #481, opened 2025-10-15
Add guards to prevent divide by zero

polunma:
**_please summarize your changes_**

### DIFF
--- a/src/mam4xx/hetfrz.hpp
+++ b/src/mam4xx/hetfrz.hpp
@@ -9,6 +9,9 @@
 #include <haero/atmosphere.hpp>
 #include <haero/math.hpp>
 
+#include <cmath>
+#include <limits>
+
 #include <mam4xx/aero_config.hpp>
 #include <mam4xx/conversions.hpp>
 #include <mam4xx/mam4_types.hpp>
@@ -170,7 +173,7 @@ Real get_reynolds_num(const Real r3lx, const Real rho_air,
       vlc_drop_adjfunc;
 
   // Reynolds number
-  const Real Re = 2 * vlc_drop * r3lx * rho_air / viscos_air;
+  const Real Re = haero::max(2 * vlc_drop * r3lx * rho_air / viscos_air, 0.0);
 
   return Re;
 }
@@ -315,7 +318,7 @@ Real get_dg0imm(const Real sigma_iw, const Real rgimm) {
 
 KOKKOS_INLINE_FUNCTION
 Real get_Aimm(const Real vwice, const Real rgimm, const Real temperature,
-              const Real dg0imm) {
+              const Real dg0imm, const Real sigma_iw) {
   constexpr Real n1 = 1e19; // number of water molecules in contact with unit
                             // area of substrate [m-2] (BAD CONSTANT)
   constexpr Real hplanck = 6.63e-34; // Planck constant (BAD CONSTANT)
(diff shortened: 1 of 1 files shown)


src/mam4xx/hetfrz.hpp
```diff
@@ -380,27 +387,37 @@ void calculate_hetfrz_contact_nucleation(
                  (bad_boltzmann * temperature)) *
       Kcoll_dust_a3 * icnlx;
 
+  // ---- dt guards and small-dt finite-difference limits ----
```
mjschmidt271: Why are we doing this? This raises some concerns for me:

1. Assuming `dt < 1e-12` is the chosen setting for a run, does it seem appropriate to effectively reduce `dt` to zero for these calculations?
    - This isn't yet to the point of losing precision, and how small would we expect `dt` to ever be?
2. This change introduces a spurious infinity in order to give the same result as a comparison we've already done.
    - This is an inappropriate way of accomplishing something much simpler, so can we outline what the goal is?
3. Also, this leads to guaranteed unnecessary calculations:
    - introducing 3 `if`s to assign the desired values of `rate_over_dt_cnt<i>` when the result is already determined by the initial `if (dt < 1e-12`).
    - introducing 3 calls to `min()` when this is also predetermined
    - Note that this all assumes that the `J` values are strictly positive, but that seems to be the case because things would look straaaange otherwise

- Now, assuming setting `dt = 0` is an appropriate choice, nearly all of the new code is unnecessary.
    - It effectively defines division by zero to be equal to infinity (it is not) and then appears to take great pains to not explicitly divide by zero.

Here, for the quantities `q<i> = {frzbccnt, frzducnt}` and in the case that their corresponding `do_<xyz>` flag is true, this appears to lead to the same result as

```c++
if (dt <= 1e-12) {
  // this will never be smaller than x2, so why do the comparison?
  x1 = z * infinity;
  x2 = z * J<i>;
  q<i> += min(x1, x2);
} else {
  // both of these should be strictly positive, and if I'm not mistaken,
  // for fixed J > 0, we always have x2 >= x1
  // so, again, introducing all of the comparisons is unnecessary
  x1 = z * ((1.0 - haero::exp(-Jcnt_bc * deltat)) / deltat);
  x2 = z / deltat;
 q<i> += min(x1, x2);
}
```

src/mam4xx/hetfrz.hpp
```diff
@@ -818,9 +877,12 @@ void hetfrz_classnuc_calc(
...
   // critical germ size
   // assume 98% RH in mixed-phase clouds (Korolev & Isaac, JAS 2006)
-  const Real rgdep = 2.0 * vwice * sigma_iv /
-                     (bad_boltzmann * temperature *
-                      haero::log(Hetfrz::rhwincloud * supersatice));
+  Real rhw_ss = Hetfrz::rhwincloud * supersatice;
+  Real rhw_ss_clamped = haero::max(rhw_ss, 1.0 + 1.0e-12);
+  Real denom = haero::max(bad_boltzmann * temperature * haero::log(rhw_ss_clamped),1.0e-30);
+  const Real rgdep = 2.0 * vwice * sigma_iv / denom;
+                     //(bad_boltzmann * temperature * haero::log(rhw_ss_clamped));
```
mjschmidt271: please remove leftover comments

src/mam4xx/hetfrz.hpp
```diff
@@ -315,17 +318,21 @@ Real get_dg0imm(const Real sigma_iw, const Real rgimm) {
 
 KOKKOS_INLINE_FUNCTION
 Real get_Aimm(const Real vwice, const Real rgimm, const Real temperature,
```
mjschmidt271: This is causing test failure for gpu. No idea why cpu still passes

Also, please delete leftover comments

src/mam4xx/hetfrz.hpp
```diff
@@ -380,27 +387,37 @@ void calculate_hetfrz_contact_nucleation(
                  (bad_boltzmann * temperature)) *
       Kcoll_dust_a3 * icnlx;
 
+  // ---- dt guards and small-dt finite-difference limits ----
+  constexpr Real eps_dt = 1e-12;
+  const bool small_dt = haero::abs(deltat) <= eps_dt;
+  const Real inv_dt   = small_dt ? std::numeric_limits<Real>::infinity()
```
odiazib: @jeff-cohere @bartgol, can we use `std::numeric_limits<Real>::infinity()` in GPUs (CUDA)? What I recall is that we cannot use standard features inside GPU kernels, but I do not know if this is okay nowadays.

PR #489, opened 2026-02-02
Prevent division by zero in the hetfrz routine.

odiazib:
We are encountering a division-by-zero error in hetfrz (see issue #490). To prevent this, we apply a small non-zero lower bound to dstcoat. The utils::min_max_bound routine then clamps the ratio value within [0.001, 1.0]:

```
for (int ispec = 0; ispec < Hetfrz::hetfrz_aer_nspec; ++ispec) {
    dstcoat[ispec] = utils::min_max_bound(0.001, 1.0, dstcoat[ispec]);
  }

```
 @susburrows

### DIFF
--- a/src/mam4xx/hetfrz.hpp
+++ b/src/mam4xx/hetfrz.hpp
@@ -1026,7 +1026,7 @@ void calculate_coated_fraction(
 
   // dust_a1
   coat_ratio1 = vol_shell[1] * (r_dust_a1 * 2.0) * fac_volsfc_dust_a1;
-  coat_ratio2 = haero::max(6.0 * dr_so4_monolayers_dust * vol_core[1], 0.0);
+  coat_ratio2 = haero::max(6.0 * dr_so4_monolayers_dust * vol_core[1], 1.0e-36);
   dstcoat[1] = coat_ratio1 / coat_ratio2;
 
   // dust_a3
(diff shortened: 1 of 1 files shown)


src/mam4xx/hetfrz.hpp
```diff
@@ -1020,14 +1020,19 @@ void calculate_coated_fraction(
 
   const Real n_so4_monolayers_dust = 1.0;
   const Real dr_so4_monolayers_dust = n_so4_monolayers_dust * 4.76e-10;
+  // NOTE: To prevent division by zero, we use the routine safe_denominator
+  // that returns a factor (1/value). In this case, value is coat_ratio2.
   Real coat_ratio2 =
-      haero::max(6.0 * dr_so4_monolayers_dust * vol_core[0], 1.0e-36);
-  dstcoat[0] = coat_ratio1 / coat_ratio2;
+      haero::max(6.0 * dr_so4_monolayers_dust * vol_core[0], 0.0);
```
susburrows: If I am reading the code correctly here, coat_ratio2 returns zero only in the case where vol_core = 0.  (The other terms in this equation are constants.)

In turn, vol_core = 0 only when dmac = 0.

dmac is the dust mass air concentration.  coat_ratio2 is used here to calculate the surface coating of the dust aerosol present in the air.  If there is no dust aerosol (dmac=0), then it makes sense for coat_ratio2 to return NaN, because the entire calculation is physically meaningless in this condition.

Because of this, I suggest the following may be a more appropriate and efficient solution: check whether dmac exceeds a minimum threshold at line 1018; if not, skip the calculations in lines 1018-1029 entirely.

Similar reasoning can be applied to all of the subsequent code blocks.

susburrows (commented):
Hi @odiazib and @kaizhangpnl: I reviewed these changes and I think there may be a better approach. Can you please check my suggestion?

src/mam4xx/hetfrz.hpp
kaizhangpnl: If vol_core[1] is used (vol_core(2) in Fortran), I think Susannah's fix is good.

PR #492, opened 2026-02-16
Fixes wet scavenging validation tests to remove erroneous biases.

jeff-cohere:
This PR contains fixes to our wet scavenging validation tests. This demonstrates that many large errors in the validation process come from defects in the testing process and not the Fortran -> C++ port, which is great news!

There are no code changes in mam4xx proper--only the validation tests themselves.


jeff-cohere:
The formatting check is broken. For what it's worth, the code is properly formatted.

PR #501, opened 2026-02-23
Undo slight optimization that was causing non-BFB results.

overfelt:


overfelt:
@odiazib Yes, once this mam4xx branch is merged to main, I'll update the E3SM branch to point to the newest version of mam4xx main.

PR #510, opened 2026-02-27
Added support for output-field-level bit-for-bit hashing.

jeff-cohere:
This feature can be enabled by placing `-f` after the input file name on the command line invocation of a validation driver. For example:

```
$ ./aero_model_driver stand_aero_model_calcsize_water_uptake_dr_ts_379.yaml -f
Kokkos::OpenMP::initialize WARNING: OMP_PROC_BIND environment variable not set
  In general, for best performance with OpenMP 4.0 or better set OMP_PROC_BIND=spread and OMP_PLACES=threads
  For best performance with OpenMP 3.1 set OMP_PROC_BIND=true
  For unit testing set OMP_PROC_BIND=false

./aero_model_driver: reading stand_aero_model_calcsize_water_uptake_dr_ts_379.yaml
./aero_model_driver: writing mam4xx_stand_aero_model_calcsize_water_uptake_dr_ts_379.py
mam4xx hash> exe=aero_model_driver input=stand_aero_model_calcsize_water_uptake_dr_ts_379.yaml date=2026-01-27-83933
      input: 5cba3a3366230ab0
 dgnumdry_m: 5549d65ce4f1eeff
   dgnumwet: 57665e9db29025b7
    ptend_q: cc2502535bb91c97
    qaerwat: 62469b2c0c8ac354
       qqcw: 43ed63a1d05c36c9
    state_q: 17908af35365a080
    wetdens: a6a1caf89328460a
```

We could use the more standard approach of placing options before the input file, but this would require changing all the drivers and I just wanted to hack something in. Willing to update if anyone feels strongly about this.

### DIFF
--- a/src/validation/validation.cpp
+++ b/src/validation/validation.cpp
@@ -88,17 +88,11 @@ void print_bfbhash(Ensemble *e) {
           std::string name;
           Real scalar;
           while (output.next_scalar(name, scalar)) {
-            if (accum_by_field.find(name) == accum_by_field.end()) {
-              accum_by_field[name] = HashType();
-            }
             mam4::bfbhash::hash(scalar, accum_by_field[name]);
           }
 
           std::vector<Real> array;
           while (output.next_array(name, array)) {
-            if (accum_by_field.find(name) == accum_by_field.end()) {
-              accum_by_field[name] = HashType();
-            }
             hash(array, accum_by_field[name]);
           }
         });


odiazib (commented):
It looks good to me. Thanks for adding hashes per variable. I will try using this feature soon. I’m okay with merging this PR.

PR #525, opened 2026-04-02
Use uniform include header guard macro names.

overfelt:
Make MAM4XX_ the initial macro name characters for the #ifndef at the top of mam4xx header files. There were only a couple of header files that did not use this mam4xx standard.


jeff-cohere (approved):
Excellent. Thanks, James!

odiazib (approved):

PR #526, opened 2026-04-02
Replace arrays of views wtih multidim views in dropmixnuc.

overfelt:
This is a simple replacement of arrays of views, e.g. `View1D raercol[pver][2] -> View3D raercol `. 

Defining as multidimensional views is more efficient in memory uses as only a single Kokkos view is used instead of many. It also allows for the memory to be allocated in a single allocation instead of many separate allocations. It also simplifies the code.  

[BFB]


src/mam4xx/ndrop.hpp
```diff
@@ -1297,14 +1297,16 @@ void update_from_explmix(
...
           const int km1 = haero::max(k - 1, top_lev);
-          const View1D &raercol_km1_nsav = raercol[km1][nsav];
-          const View1D &raercol_k_nsav = raercol[k][nsav];
-          const View1D &raercol_kp1_nsav = raercol[kp1][nsav];
-          const View1D &raercol_k_nnew = raercol[k][nnew];
-          const View1D &raercol_cw_km1_nsav = raercol_cw[km1][nsav];
-          const View1D &raercol_cw_k_nsav = raercol_cw[k][nsav];
-          const View1D &raercol_cw_kp1_nsav = raercol_cw[kp1][nsav];
-          const View1D &raercol_cw_k_nnew = raercol_cw[k][nnew];
+          const View1D raercol_km1_nsav = ekat::subview(raercol, km1, nsav);
```
odiazib: raercol in the EAMxx interface will have the layout `(ncol, nlev, nsav)`. Does it make a difference if we use `(ncol, nsav, nlev)`, particularly in terms of performance and memory access? In our case, the fastest-varying index in memory will be `nsav`. With the layout `(ncol, nlev, nsav)`, will accesses along nsav be contiguous?

odiazib (approved):

jeff-cohere (approved):

PR #528, opened 2026-04-07
Testing out new EKAT install in standalone mam4xx for Luca.

jeff-cohere:
EKAT is switching to use FetchContent for some of its dependencies. This PR is just a test of these changes to EKAT's installation procedures to make sure it still works for mam4xx. It won't be merged.


bartgol:
After modifying the workflow file to tell actions/checkout to grab the correct ekat branch, this seems to work. We can close.

PR #532, opened 2026-04-10
Moves the AeroProcess class (unused by EAMxx) to the validation suite.

jeff-cohere:
References to AeroProcess and related type aliases are removed. Unit tests that only test AeroProcess instances are also removed. Existing validation tests that use AeroProcess retain access to this class.

Closes #522


src/tests/CMakeLists.txt
```diff
@@ -23,14 +23,10 @@ target_link_libraries(mam4xx_tests mam4xx)
 
 EkatCreateUnitTest(utils_unit_tests utils_unit_tests.cpp
   LIBS mam4xx_tests USER_DEFINED_TEST_SESSION)
-EkatCreateUnitTest(mam4_nucleation_unit_tests mam4_nucleation_unit_tests.cpp
```
odiazib, line 26: We may want to check whether this unit test is exercising code that we currently use in EAMxx.

src/tests/CMakeLists.txt
```diff
@@ -23,14 +23,10 @@ target_link_libraries(mam4xx_tests mam4xx)
...
-  LIBS mam4xx_tests USER_DEFINED_TEST_SESSION)
 EkatCreateUnitTest(mam4_gasaerexch_unit_tests mam4_gasaerexch_unit_tests.cpp
   LIBS mam4xx_tests USER_DEFINED_TEST_SESSION)
 EkatCreateUnitTest(mam4_coagulation_unit_tests mam4_coagulation_unit_tests.cpp
   LIBS mam4xx_tests USER_DEFINED_TEST_SESSION)
-EkatCreateUnitTest(mam4_calcsize_unit_tests mam4_calcsize_unit_tests.cpp
```
odiazib, line 32: We may want to check whether this unit test is exercising code that we currently use in EAMxx.

src/tests/CMakeLists.txt
jeff-cohere: This test just constructs a NucleationProcess class and calls `compute_tendencies()` on it. I don't think we use either of these things.

src/tests/CMakeLists.txt
jeff-cohere: Like the nucleation test, this just constructs a CalcSizeProcess and computes tendencies. In both this and the nucleation test, all we do is check that the tendencies are not NaNs.

odiazib (approved):
Thanks for cleaning this up. I wish we could use this framework in EAMxx. However, the EAMxx team already has an established framework, which makes this approach difficult to use there.

PR #539, opened 2026-04-17
Replacing workspace array of views with a 2D view in wet scavenging code.

jeff-cohere:
In line with James's recent slew of optimizations, I'm replacing an array of scratch views with a single 2D view. Maybe we'll see a little performance boost?

### DIFF
--- a/src/mam4xx/wet_dep.hpp
+++ b/src/mam4xx/wet_dep.hpp
@@ -1970,7 +1970,8 @@ void aero_model_wetdep(
   work_ptr += aero_model::pcnst * 2;
 
   const int num_tendencies_workspaces = 14;
-  View2D compute_q_tendencies_workspace(work_ptr, num_tendencies_workspaces, mam4::nlev);
+  View2D compute_q_tendencies_workspace(work_ptr, num_tendencies_workspaces,
+                                        mam4::nlev);
   work_ptr += num_tendencies_workspaces * mam4::nlev;
 
   /// error check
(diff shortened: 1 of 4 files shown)


jeff-cohere:
To be clear: I don't expect much of a performance gain from this, since the memory was already being allocated from a contiguous block. But this change cleans things up a bit and gets this "array of views" pattern off our radar.

overfelt (approved):

src/mam4xx/wet_dep.hpp
```diff
@@ -1165,18 +1165,19 @@ void compute_q_tendencies(
...
     evap[k] = mam4::max(0.0, evapr[k] * pdel[k] / gravit);
     rain[k] = mam4::max(0.0, prain[k] * pdel[k] / gravit);
   });
 
-  View1D precabs = workspace[3];
-  View1D precabs_tmp = workspace[4];
-  View1D precabs_base = workspace[5];
-  View1D precabs_base_tmp = workspace[6];
+  // (look below for workspace 2)
+  View1D precabs = ekat::subview(workspace, 3);
```
odiazib, line 1172: I wonder if we should use enums, similar to what James did for Hetfrz. In my experience, using numbers can lead to bugs.

jeff-cohere: I think that's a good idea where practical, but here the work arrays are reused for several purposes, so I don't see how we can assign meaningful names to these indices.

jeff-cohere: The good news is that each such "temporary" work array component is used within a pretty tight scope, so I'm not too concerned about clarity in this case.

odiazib:
Before merging, can we open a PR in E3SM so we can test that it doesn’t break any tests there?

PR #530, opened 2026-04-10
Removing single loops for photo table.

odiazib:
It removes two single loops from the photo table computation.
Before PR
image
After PR
image

It closes https://github.com/eagles-project/mam4xx/issues/513

PR #540, opened 2026-04-21
Rewrite gas_washout function to eliminate Kokkos::single. 

overfelt:
There was a loop over the column that used a Kokkos::single to iterate.  With a little refactoring of the code this was replaced with a Kokkos::parallel.

Performance testing of this kernel demonstrated a 3-4x speedup with these changes, as shown here:

image

But since this kernel was not expensive to begin with, the overall speedup of the aerosol processes was not significant:

image (1)

Note that this kernel is part of the aero_microphysics timing.

### DIFF
--- a/src/mam4xx/mo_sethet.hpp
+++ b/src/mam4xx/mo_sethet.hpp
@@ -131,30 +131,31 @@ void gas_washout(
     allca[kk] = 0; // pretend allca is a kokkos view.
   });
   for (int k = ktop; k < pver; k++) {
-    Kokkos::parallel_for(Kokkos::TeamVectorRange(team, ktop, k+1), [&](int kk) {
-      if (rain(kk) != 0.0) { // finding rain cloud
-        const Real xeqca =
-            xgas(k) /
-            (xliq(kk) * avo2 + 1.0 / (xhen_i(k) * const0 * tfld_i(k))) *
-            xliq(kk) * avo2;
-        //-----------------------------------------------------------------
-        //       ... calculate ca; inside cloud concentration in  #/cm3(air)
-        //-----------------------------------------------------------------
-        const Real xca = geo_fac * xkgm * xgas(k) / (xrm * xum) * delz_i(k) *
-                         xliq(kk) * cm3_2_m3;
+    Kokkos::parallel_for(
+        Kokkos::TeamVectorRange(team, ktop, k + 1), [&](int kk) {
+          if (rain(kk) != 0.0) { // finding rain cloud
+            const Real xeqca =
+                xgas(k) /
+                (xliq(kk) * avo2 + 1.0 / (xhen_i(k) * const0 * tfld_i(k))) *
+                xliq(kk) * avo2;
+            //-----------------------------------------------------------------
+            //       ... calculate ca; inside cloud concentration in  #/cm3(air)
+            //-----------------------------------------------------------------
+            const Real xca = geo_fac * xkgm * xgas(k) / (xrm * xum) *
+                             delz_i(k) * xliq(kk) * cm3_2_m3;
 
-        // -----------------------------------------------------------------
-        //       ... if is not saturated (take hno3 as an example)
-        //               hno3(gas)_new = hno3(gas)_old - hno3(h2o)
-        //           otherwise
-        //               hno3(gas)_new = hno3(gas)_old
-        // -----------------------------------------------------------------
-        allca[kk] += xca;
-        if (allca[kk] < xeqca) {
-          xgas(k) = mam4::max(xgas(k) - xca, 0.0);
-        }
-      }
-    });
+            // -----------------------------------------------------------------
+            //       ... if is not saturated (take hno3 as an example)
+            //               hno3(gas)_new = hno3(gas)_old - hno3(h2o)
+            //           otherwise
+            //               hno3(gas)_new = hno3(gas)_old
+            // -----------------------------------------------------------------
+            allca[kk] += xca;
+            if (allca[kk] < xeqca) {
+              xgas(k) = mam4::max(xgas(k) - xca, 0.0);
+            }
+          }
+        });
     team.team_barrier();
   }
 } // end subroutine gas_washout
--- a/src/mam4xx/mo_sethet.hpp
+++ b/src/mam4xx/mo_sethet.hpp
@@ -131,7 +131,7 @@ void gas_washout(
     allca[kk] = 0; // pretend allca is a kokkos view.
   });
   for (int k = ktop; k < pver; k++) {
-    Kokkos::parallel_for(Kokkos::TeamVectorRange(team, k, pver), [&](int kk) {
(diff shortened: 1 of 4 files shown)


src/validation/mo_sethet/gas_washout.cpp
```diff
@@ -40,8 +40,9 @@ void gas_washout(Ensemble *ensemble) {
     Kokkos::parallel_for(
         team_policy, KOKKOS_LAMBDA(const mam4::ThreadTeam &team) {
           Kokkos::single(Kokkos::PerTeam(team), [=]() {
-            gas_washout(team, plev - 1, xkgm, xliq_ik, xhen_i, tfld_i, delz_i,
-                        xgas);
+	      
+//          gas_washout(team, 0, 1, xkgm, xliq_ik, xhen_i, tfld_i, delz_i,
```
jeff-cohere: Should we remove this `Kokkos::single` and just call `gas_washout` in its place?

src/mam4xx/mo_sethet.hpp
```diff
@@ -126,32 +128,35 @@ void gas_washout(
...
-    allca += xca;
-    if (allca < xeqca) {
-      xgas(k) = mam4::max(xgas(k) - xca, 0.0);
-    }
+  Real allca[mam4::nlev] = {};
+  Kokkos::parallel_for(Kokkos::TeamVectorRange(team, pver), [&](int kk) {
+    allca[kk] = 0; // pretend allca is a kokkos view.
+  });
+  for (int k = ktop; k < pver; k++) {
+    Kokkos::parallel_for(Kokkos::TeamVectorRange(team, k, pver), [&](int kk) {
```
jeff-cohere: Ahhh, I overlooked this opportunity!

overfelt: I'm not positive this is equivalent to the original code.  I think it is, but I wonder if that allca array is used correctly.

src/validation/mo_sethet/gas_washout.cpp
overfelt: With the outer loop pushed into the gas_washout function, I don't think I can get the same numbers as before out of the function in order to make this validation test pass.   Maybe update the input yaml file with a new solution?

src/mam4xx/mo_sethet.hpp
jeff-cohere: Indeed. Right now I notice that the original `allca` is purely accumulative and is never reset to 0, whereas yours accumulates for each `kk` separately. In that sense, I think it might be different, unless I'm missing something. We could get the validation test working to give some more conclusive feedback.

jeff-cohere: But then again, you've pulled the loop inside `gas_washout`. Yes, I think it's worth working up a new validation test to see what this is doing.

src/validation/mo_sethet/gas_washout.cpp
jeff-cohere: If we stick the old `gas_washout` implementation into the validation test and call it in a manner congruent to the new one, we should be able to compare the output of both. I'm happy to try to put this together if you like.

overfelt: After spending hours trying to come up with a way to avoid the complex bottleneck in this function, I'm suspicious that the fix turned out to be so simple. 

Yes you are correct, unit testing with a single example might miss some cases and this is simple enough to copy the old implementation to a test and run it to see if the new implementation is correct. 

There is also the need to make that scratch array allocated as a scratch view in e3sm and passed in instead of a local static array.

src/mam4xx/mo_sethet.hpp
```diff
@@ -126,32 +128,35 @@ void gas_washout(
...
-    //               hno3(gas)_new = hno3(gas)_old - hno3(h2o)
-    //           otherwise
-    //               hno3(gas)_new = hno3(gas)_old
-    // -----------------------------------------------------------------
-    allca += xca;
-    if (allca < xeqca) {
-      xgas(k) = mam4::max(xgas(k) - xca, 0.0);
-    }
+  Kokkos::parallel_for(Kokkos::TeamVectorRange(team, pver),
+                       [&](int kk) { scratch[kk] = 0; });
```
odiazib, line 132: One thing I heard in the Kokkos tutorial is that it is recommended to use [=] instead of [&]. I realize that capturing by value can increase overhead, but it doesn’t seem that bad. I am just curious whether @jeff-cohere or @overfelt have any comments on this.

src/mam4xx/mo_sethet.hpp
```diff
@@ -126,32 +128,35 @@ void gas_washout(
...
+  for (int k = ktop; k < pver; k++) {
```
odiazib, line 133: Does it make a difference if the loop is converted to a parallel_for instead of the one on line 134?

odiazib (approved):
It looks great! I just have a couple of questions that don’t require any code changes.

src/mam4xx/mo_sethet.hpp
jeff-cohere: Capture by value is always easier to reason about, and it's likely the only thing possible when moving data between a host and device. But I'm not sure if it matters in this case. Curious what @overfelt thinks.

src/mam4xx/mo_sethet.hpp
jeff-cohere: Yes, because we're storing data in `xgas` that is updated sequentially (and indexing it with `k`).

jeff-cohere (approved):
Thanks for putting this together, James. Yours was a good hunch. I'm curious to see how much this changes answers (given that the BFB hashes differ).

src/mam4xx/mo_sethet.hpp
overfelt: Yes, when lambda capturing to copy to device, capture by value makes sense since everything is a copy anyway. 

Mam4xx uses capture by reference for all the device level TeamVectorRange parallel_for loops but I can not find an example where it would change the result if capture by value was used instead.  I think of capture by reference being used when the intent is to update the values captured to the lambda, which is the intent here. Capture by value is when the intent is not to change the value in the owning scope.  Since it is a Kokkos view being changed, it does not matter which is used but I still like the showing of intent.  

Did the Kokkos tutorial talk about device level lambdas?  Or was it about host-to-device lambdas decorated with KOKKOS_LAMBDA?

overfelt:
image

overfelt:
image (1)

odiazib:
Nice to see a 3× performance improvement. It’s also great that we were able to remove part of the code that could cause issues in the future. Nice work, James!

odiazib:
The aero–microphysics interface accounts for 11.8% of the total MAM4xx cost, which corresponds to 5.9% of the total cost. Within this interface, the most expensive kernel is `modal_aero_amicphys_intr` (see PR #530). This kernel is very complex, but I don’t see any internal parallelism or nested parallelism. We could look at this kernel next.

PR #542, opened 2026-04-24
It appears that mo_chm_diags.hpp is not used outside of some validation tests.

overfelt:
Can we delete mo_chm_diags.hpp? It is not used anywhere but maybe we are keeping it for future use?


odiazib:
> Can we delete mo_chm_diags.hpp? It is not used anywhere but maybe we are keeping it for future use?

I thought we were using this code in the aero-microphysics interface. I vote for removing it, but I would like to hear from @mingxuanwupnnl or @kaizhangpnnl.

mingxuanwupnnl:
mo_chm_diags.hpp is just for computing diagnostics for bulk quantities, like bulk mass concentrations for bc. I would suggest keeping it for now. We can exclude it for code compling.

PR #544, opened 2026-04-28
Bringing in Meng's Kokkos kernel plotting scripts.

jeff-cohere:
These scripts parse and plot output from the kokkos-tools kernel timers. This PR only brings them under source control. I think we can generalize them in a subsequent PR.

### DIFF
--- a/scripts/make_kernel_plots.sh
+++ b/scripts/make_kernel_plots.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+# NOTE: This script only works on Perlmutter at the moment, and relies on tools
+# NOTE: built in Meng's directory.
+
+case_dir=/pscratch/sd/m/meng/scream_scratch/SMS_P32x1_Ld10.ne30pg2_oECv3.F2010-EAMxx-MAM4xx.pm-gpu_gnugpu.ThirdTimingViews/run
+plotdir=/pscratch/sd/m/meng/tmp/kokkos_tools
+exe_kk="/global/u1/m/meng/utils/kokkos-tools/profiling/simple-kernel-timer/kp_json_writer"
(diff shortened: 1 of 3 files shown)


overfelt (approved):

scripts/make_kernel_plots.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+# NOTE: This script only works on Perlmutter at the moment, and relies on tools
+# NOTE: built in Meng's directory.
+
+case_dir=/pscratch/sd/m/meng/scream_scratch/SMS_P32x1_Ld10.ne30pg2_oECv3.F2010-EAMxx-MAM4xx.pm-gpu_gnugpu.ThirdTimingViews/run
+plotdir=/pscratch/sd/m/meng/tmp/kokkos_tools
+exe_kk="/global/u1/m/meng/utils/kokkos-tools/profiling/simple-kernel-timer/kp_json_writer"
```
overfelt, line 8: Maybe this could be copied to somewhere inside of /pscratch/sd/m/meng so it is accessable withou opening up permissions on the home directory.

jeff-cohere: Maybe so, but I was just copying it here so I can get it to work on Aurora. :-)

jeff-cohere: At length, I would like for us to incorporate kokkos-tools (the kernel timers, at the very least) into E3SM's build as an optional feature. I'm talking with Noel about this and may just do it if provoked!

jeff-cohere: See [EKAT PR 436](https://github.com/E3SM-Project/EKAT/pull/436).

PR #547, opened 2026-04-30
Replacing static global C arrays with Kokkos::Arrays

### LINKED ISSUE #364: Delete the duplicate of adv_mass.
See here.
https://github.com/eagles-project/mam4xx/blob/b43b50b645ca80584ca05cc66f89279a1643ef99/src/mam4xx/gas_chem_mechanism.hpp#L34

jeff-cohere:
Here's a cool pattern motivated by #364. It might be useful elsewhere in mam4xx and perhaps even EAMxx.

Basically, I've just replaced some C arrays with Kokkos::Arrays of the appropriate type. These are all `constexpr` and can be brought onto device with minimal effort.

Closes #364


jeff-cohere:
Fyi @tcclevenger

src/mam4xx/gas_chem_mechanism.hpp
```diff
@@ -22,34 +22,20 @@ constexpr int extcnt = 9;     // number of species with external forcing
...
-      47.998200,     34.013600,  98.078400,     64.064800, 62.132400,
-      12.011000,     115.107340, 12.011000,     12.011000, 12.011000,
-      135.064039,    58.442468,  250092.672000, 1.007400,  115.107340,
-      12.011000,     58.442468,  250092.672000, 1.007400,  135.064039,
-      58.442468,     115.107340, 12.011000,     12.011000, 12.011000,
-      250092.672000, 1.007400,   12.011000,     12.011000, 250092.672000,
-      1.007400};
-  return adv_mass;
-}
+constexpr auto permute_4 = Kokkos::to_array<int>(
```
odiazib, line 25: Will using `Kokkos::to_array `improve computational performance? We use many C static arrays in the ACI interface; should we remove those as well?

see here: https://github.com/E3SM-Project/E3SM/blob/411ca466d80070de825f6e0f67ee1e01d7007e6c/components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp#L284
and here:
https://github.com/eagles-project/mam4xx/blob/3963e21d5085abec8041a923058cdf55a9dfe9b1/src/mam4xx/ndrop.hpp#L62

odiazib (approved):

src/mam4xx/gas_chem_mechanism.hpp
jeff-cohere: I don't know how much these arrays improve computational performance, but they're definitely a simpler representation of array data that can be used on-device. One limitation is that `Kokkos::Array` can only be used for 1D arrays, so I suspect we wouldn't be able to replace certain other array data without flattening them.

overfelt: Like std::array, Kokkos::Array is a little more intuitive when passing as a function argument.  Just have to make sure to pass by reference to avoid the copy.  Like std::array, I assume Kokkos::Array can be nested to make 2D like structures.  The syntax is long,  but maybe the correct `using` statement would help, something like 
`   
template <class X, int n, int m> using Array2D = Kokkos::Array<Kokkos::Array<X,m>,n>;
`
so
```
  const Array2D<Real, 3,4> X={1.1, 1.2, 1.3,1.4, 2.1, 2.2, 2.3, 2.4, 3.1, 3.2, 3.3, 3.4}; 
  Y = 2*X[2][3];
```

overfelt (approved):

Issue #557, opened 2026-05-21
Make aerosol species hygroscopicity a specifiable parameter


jeff-cohere:
It was pointed out at our last eval meeting that we need to be able to set POM hygroscopicity for tuning. Currently, all of our aerosol species have hard-coded hygroscopicities. We can keep the hard-coded values as defaults, but we need to be able to accept user-specified values. This new capability requires a PR for mam4xx and another for EAMxx (for parsing input YAML files for any user-specified values).

jeff-cohere:
Hmm. There are a few places where these hygroscopicities are defined because of legacy spaghetti. I am going to attempt to do some light reorganization to attempt to thread these into a single source of truth.

jeff-cohere:
This issue isn't quite completed yet. The remaining work is to insert the namelist parameters for the tunable aerosol species properties on the EAMxx side.

PR #560, opened 2026-06-15
Code cleanup and changes related to #557.

jeff-cohere:
Specifically:

* Functions related to aerosol species have been moved to `aero_species.hpp` and `aero_species.cpp`.
* The `mam4::aero_species` function now accepts a `mam4::AeroId` instead of an integer aerosol index.
* The `mam4::aero_species` function now refers to a persistent set of mutable aerosol species whose properties can be overridden.
* Functions have been added for overriding aerosol species properties.
* The change in the `mam4::aero_species` function signature allowed me to identify several aerosol property indexing bugs which I've attempted to fix. These fixes need reviewing by the mam4xx team.

## TODO

Global views are not supported on GPUs, so we can't call `aero_species` to get properties.

* Identify code locations using `aero_species` function
* Make sure a device view of `AeroSpecies` is available in each of these contexts.

**It's likely that I'll close this PR in favor of a more appropriate solution.**

Fixes #558


src/mam4xx/aero_species.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+void set_aero_density(const AeroId id, Real density) {
+  aero_species_[int(id)].density = density;
+}
+
+void set_aero_hygroscopicity(const AeroId id, Real hygroscopicity) {
+  aero_species_[int(id)].hygroscopicity = hygroscopicity;
+}
+
+std::string aero_id_str(const AeroId id) {
```
odiazib, line 82: We must consolidate the species names and other properties given in the array with the hard-coded information in [mam_coupling.hpp](https://github.com/E3SM-Project/E3SM/blob/35736e3614877a248926f6f18b617e0466ecf8f0/components/eamxx/src/physics/mam/mam_coupling.hpp#L89).

odiazib (changes requested):
Thanks, @jeff-cohere, for cleaning up this part of the code. We have hard-coded species indices and names in mam_coupling.hpp. Have you looked into whether we can consolidate the changes in this PR with mam_coupling? Keeping species names, properties, and related information in one place will help when we decide to adopt another reaction mechanism or add a new mode.

Before merging this PR, we should open a PR in E3SM, and if possible, run a 10-day case to make sure performance is not impacted.

PR #561, opened 2026-07-07
Attempt #2 at making aerosol species properties configurable.

### LINKED ISSUE #557: Make aerosol species hygroscopicity a specifiable parameter

jeff-cohere:
This PR creates a couple of new View types for storing aerosol species properties, and inserts them into the functions that need them currently. It would be good for us to reorganize things so that we pass this data around more systematically--what we call "mam4xx" is currently just a big pile of functions with many, many undocumented assumptions.

Because we typically store aerosol processes on "the device" (usually but not always the GPU), I had to change the way that some of the tests were written in order to either run them on the device or make the functions accept both host- and device-stored aerosol properties. See the discussion in #557 for details.

Closes #557


src/mam4xx/modal_aero_calcsize.hpp
```diff
@@ -18,8 +18,13 @@ namespace modal_aero_calcsize {
 
 constexpr int maxd_aspectype = ndrop::maxd_aspectype;
 
+// NOTE: init_calcsize is only called from the CalcsizeData's init method, so we
+// can use host-side NOTE: aerosol species properties. We should really get rid
+// of these class-like structures; they NOTE: don't really work well on GPUs.
```
overfelt: Clang formatter messed up comment.

src/mam4xx/modal_aero_calcsize.hpp
```diff
@@ -18,8 +18,13 @@ namespace modal_aero_calcsize {
...
+// -JNJ, 2026-07-07
 KOKKOS_INLINE_FUNCTION
 void init_calcsize(
```
overfelt: Is KOKKOS_INLINE_FUNCTION needed if using AeroSpeciesHostView does not allow to be called from device.

overfelt (approved):

meng630 (approved):

src/mam4xx/aero_species.hpp
```diff
@@ -7,21 +7,95 @@
...
+
+// default values for aerosol species properties
+namespace defaults {
+
+/// Molecular weight of mam4 dust aerosol [kg/mol]
+static constexpr Real mam4_molec_weight_dst = 0.135065;
+
+/// Molecular weight of mam4 marine organic matter [kg/mol]
+static constexpr Real mam4_molec_weight_mom = 250.093;
+
```
mingxuanwupnnl, line 58: @jeff-cohere would you need to add molecular weight for other species as well?

jeff-cohere: Possibly! This is just a constant that I moved from elsewhere.

PR #563, opened 2026-07-14
add scale factors for dust and seasalt emission

mingxuanwupnnl:
Current MAMxx uses hard-wired emission factors, this PR add two scale factors

srf_emis_scale_factor_for_seasalt (default value is 0.6)
srf_emis_scale_factor_for_dust (default value is 1.5)

as namelist variables for dust and seasalt emission in MAMxx. This PR should be BFB.


kaizhangpnl (approved):
The revised code looks good to me. The only thing I noticed is that this change may create an inconsistency with the validation test code. Will this fail the integration tests?

jeff-cohere (approved):
Looks good!

PR #565, opened 2026-07-16
Fixed some aerosol species indexing problems reported in #558.

jeff-cohere:
This PR fixes some aerosol indices that are improperly used to fetch aerosol species properties. I noticed a pattern in which we mistakenly use the offset of an aerosol species within its mode to refer to a specific aerosol species. One might call this "using an aerosol index (position 0, 1, 2, within a mode) in place of an aerosol identifier (SO4, SOA, POM). See #558 for details.

Closes #558

### DIFF
--- a/src/mam4xx/mode_wet_particle_size.hpp
+++ b/src/mam4xx/mode_wet_particle_size.hpp
@@ -218,13 +218,13 @@ void mode_avg_wet_particle_diam_water_uptake(const Diagnostics &diags,
 //    @param dgn_awet [out] geometric mean diameter (m) of each aerosol mode
 // ------------------------------------------------------------------------
 KOKKOS_INLINE_FUNCTION
-void diag_dgn_wet(
-    const mam4::AeroConfig &aero_config,
-    const Real qaer_cur[mam4::AeroConfig::num_aerosol_ids()]
-                       [mam4::AeroConfig::num_modes()],
-    const Real qnum_cur[mam4::AeroConfig::num_modes()],
-    const Real molecular_weight_gm[mam4::AeroConfig::num_aerosol_ids()],
-    const Real dwet_ddry_ratio, Real dgn_awet[mam4::AeroConfig::num_modes()]) {
+void diag_dgn_wet(const mam4::AeroConfig &aero_config,
+                  const Real qaer_cur[mam4::AeroConfig::num_aerosol_ids()]
+                                     [mam4::AeroConfig::num_modes()],
+                  const Real qnum_cur[mam4::AeroConfig::num_modes()],
+                  const AeroSpeciesData<Real> molecular_weight_gm,
+                  const Real dwet_ddry_ratio,
+                  Real dgn_awet[mam4::AeroConfig::num_modes()]) {
   static constexpr int num_aer = mam4::AeroConfig::num_aerosol_ids();
   static constexpr int num_modes = mam4::AeroConfig::num_modes();
   // --------------------------
@@ -236,9 +236,9 @@ void diag_dgn_wet(
     for (int iaer = 0; iaer < num_aer; ++iaer) {
       const AeroId aid = mode_aero_species(n, iaer);
       if (aid != AeroId::None) {
-        const Real weight_gm_per_mol = molecular_weight_gm[iaer];
+        const Real weight_gm_per_mol = molecular_weight_gm[aid];
         const Real tmpa = qaer_cur[iaer][n] * weight_gm_per_mol;
-        tmp_dryvol += tmpa / aero_config.aero_species(int(aid)).density;
+        tmp_dryvol += tmpa / aero_config.aero_species[aid].density;
       }
     }
     // Convert dry volume to dry diameter, then to wet diameter
--- a/src/mam4xx/mode_wet_particle_size.hpp
+++ b/src/mam4xx/mode_wet_particle_size.hpp
@@ -234,9 +234,12 @@ void diag_dgn_wet(
     Real tmp_dryvol = 0.0;
     // Sum up the volume of all species in this mode
     for (int iaer = 0; iaer < num_aer; ++iaer) {
-      const Real weight_gm_per_mol = molecular_weight_gm[iaer];
-      const Real tmpa = qaer_cur[iaer][n] * weight_gm_per_mol;
-      tmp_dryvol += tmpa / aero_config.aero_species(iaer).density;
+      const AeroId aid = mode_aero_species(n, iaer);
+      if (aid != AeroId::None) {
+        const Real weight_gm_per_mol = molecular_weight_gm[iaer];
+        const Real tmpa = qaer_cur[iaer][n] * weight_gm_per_mol;
+        tmp_dryvol += tmpa / aero_config.aero_species(int(aid)).density;
+      }
     }
     // Convert dry volume to dry diameter, then to wet diameter
     const Real sx = std::log(mam4::modes(n).mean_std_dev);
--- a/src/mam4xx/rename.hpp
+++ b/src/mam4xx/rename.hpp
@@ -673,18 +673,20 @@ public:
           rename_idx = _mam4xx2rename_idx[imode][jspec];
 
           const AeroId aero_id = mode_aero_species(imode, jspec);
-          const Real molecular_weight =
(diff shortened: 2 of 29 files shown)


jeff-cohere:
This issue wasn't found by our tests because our tests have the same indexing error! Nor were they triggered by our validation tests, because the answers don't seem to change very much. All this means to me is that we need better tests. :-)

src/mam4xx/mode_wet_particle_size.hpp
```diff
@@ -234,9 +234,12 @@ void diag_dgn_wet(
     Real tmp_dryvol = 0.0;
     // Sum up the volume of all species in this mode
     for (int iaer = 0; iaer < num_aer; ++iaer) {
-      const Real weight_gm_per_mol = molecular_weight_gm[iaer];
-      const Real tmpa = qaer_cur[iaer][n] * weight_gm_per_mol;
-      tmp_dryvol += tmpa / aero_config.aero_species(iaer).density;
+      const AeroId aid = mode_aero_species(n, iaer);
```
kaizhangpnl, line 237: This line 

"const AeroId aid = mode_aero_species(n, iaer);" 

It seems `mode_aero_species` expects local index in that mode (https://github.com/eagles-project/mam4xx/blob/87212ee836cdb335b535c66fe2ae3d8c55a90148/src/mam4xx/aero_modes.hpp#L126), but `iaer` is the canonical species id?

jeff-cohere: It's backwards. We definitely need to improve how we refer to species within modes, because there are multiple ways of referring to things, and it's confusing.

In that loop, `iaer` is a modal species index that goes from 0 to 7 (the maximum number of aerosol species that can occupy a mode, and *coincidentally* the number of aerosol species in MAM4). Meanwhile,`aid` is actually a canonical aerosol species ID that refers to the actual species that has position `iaer` in that mode. If you want the position of a canonical species within a mode, you can call `aerosol_index_for_mode` instead.

This is a design issue. It's really difficult to navigate these situations currently.

src/mam4xx/rename.hpp
```diff
@@ -671,19 +671,20 @@ class Rename {
         for (int jspec = 0; jspec < nspec; ++jspec) {
           // get the mapping from the mam4xx species ordering to rename's
           rename_idx = _mam4xx2rename_idx[imode][jspec];
+
+          const AeroId aero_id = mode_aero_species(imode, jspec);
```
kaizhangpnl, line 694: It seems `mode_aero_species` could return `None` for unused mode/species positions, will this cause problems?

jeff-cohere: If its value is not checked, yes. I think we need to improve the way we refer to aerosol species within modes. We are talking about doing some code reorganization in the near term, and I think re-examining mode/species indexing should be a high priority.

jeff-cohere: I can add a check here. I don't know what we can assume about `nspec` in this code. Good catch.

src/mam4xx/mode_wet_particle_size.hpp
kaizhangpnl: @jeff-cohere thanks for your prompt reply! I mainly concern about the local index in the mode vs the canonical index. For example, for the accumulation mode, with iaer == 0, SOA mass and molecular weight are used, but mode_aero_species(0, 0) returns SO4 for the density?

jeff-cohere: That's curious. Thanks for looking into it, Kai. It's possible that this stuff isn't actually set up the way it's implied to be. It's also possible that the original code happens to be correct because how things happen to be. I'll take a closer look at this and try to figure out what exactly is going on.

jeff-cohere: > @jeff-cohere thanks for your prompt reply! I mainly concern about the local index in the mode vs the canonical index. For example, for the accumulation mode, with iaer == 0, SOA mass and molecular weight are used, but mode_aero_species(0, 0) returns SO4 for the density?

Looking back at this:

Here are the aerosol species orderings within the different modes in mam4xx, taken from `aero_modes.hpp`:

Mode 0 (accumulation): SO4, POM, SOA, BC, DST, NaCL, MOM
Mode 1 (aitken): SO4, SOA, NaCl, MOM
Mode 2 (coarse): DST, NaCl, SO4, BC, POM, SOA, MOM
Mode 3 (primary carbon): POM, BC, MOM

These orderings are the same as in the Fortran code for the case of MODAL_AERO_4MODE_MOM. (For MODAL_AERO_4MODE, MOM doesn't appear in these modes.) According to this information, I would expect `mode_aero_species(0, 0)` to return SO4 and `mode_aero_species(0, 2)` to return SOA. Can you explain what you mean by "for the accumulation mode, with iaer == 0, SOA mass and molecular weight are used"?

jeff-cohere: Looking at this more closely, I'm curious why we're passing molecular weights separately from the aerosol species information. This could just be leftover code, and it makes things more confusing. But looking *even more closely*, I see that the function `diag_dgn_wet` is not called by mam4xx itself OR in our validation tests OR by EAMxx. So should we get rid of it?

src/mam4xx/rename.hpp
```diff
@@ -580,21 +580,21 @@ class Rename {
...
-    _mass_2_vol[iaer_so4] =
-        config_._molecular_weight_so4 / aero_species_h(iaer_so4).density;
-    _mass_2_vol[iaer_pom] =
-        config_._molecular_weight_pom / aero_species_h(iaer_pom).density;
+    _mass_2_vol[soa_idx] =
+        config_._molecular_weight_soa / aero_species_h(soa_idx).density;
+    _mass_2_vol[so4_idx] =
+        config_._molecular_weight_so4 / aero_species_h(so4_idx).density;
+    _mass_2_vol[pom_idx] =
+        config_._molecular_weight_pom / aero_species_h(pom_idx).density;
```
jeff-cohere: This `mass_2_vol` business is confusing and another source of error, so I'm going to replace it with a pair of functions that convert between mass mixing ratios and ~~volume mixing ratios~~"volume units" $[\texttt{m}^3/\texttt{kmol air}]$ for a given aerosol species.

Is there a term for quantities expressed in these volume-per-mass ratios? Some sort of "aerosol specific volume"?

src/mam4xx/mode_wet_particle_size.hpp
kaizhangpnl: > > @jeff-cohere thanks for your prompt reply! I mainly concern about the local index in the mode vs the canonical index. For example, for the accumulation mode, with iaer == 0, SOA mass and molecular weight are used, but mode_aero_species(0, 0) returns SO4 for the density?
> 
> Looking back at this:
> 
> Here are the aerosol species orderings within the different modes in mam4xx, taken from `aero_modes.hpp`:
> 
> Mode 0 (accumulation): SO4, POM, SOA, BC, DST, NaCL, MOM Mode 1 (aitken): SO4, SOA, NaCl, MOM Mode 2 (coarse): DST, NaCl, SO4, BC, POM, SOA, MOM Mode 3 (primary carbon): POM, BC, MOM
> 
> These orderings are the same as in the Fortran code for the case of MODAL_AERO_4MODE_MOM. (For MODAL_AERO_4MODE, MOM doesn't appear in these modes.) According to this information, I would expect `mode_aero_species(0, 0)` to return SO4 and `mode_aero_species(0, 2)` to return SOA. Can you explain what you mean by "for the accumulation mode, with iaer == 0, SOA mass and molecular weight are used"?

Sorry for the delayed reply! Thanks for the note. You are right. I mistakenly took iaer as the canonical index.

kaizhangpnl: > Looking at this more closely, I'm curious why we're passing molecular weights separately from the aerosol species information. This could just be leftover code, and it makes things more confusing. But looking _even more closely_, I see that the function `diag_dgn_wet` is not called by mam4xx itself OR in our validation tests OR by EAMxx. So should we get rid of it?

I guess this is for aerosol wet size (per mode) diagnostic output. I think it's better to retain the diagnostics for the aerosol wet radius and hygroscopicity for each mode.

PR #568, opened 2026-07-22
Change how vertical transport is handled when aero_vertical_mix is false.

overfelt:
This issue is described in detail on the github issue tracker, ticket
566.  See: "enable_aero_vertical_mix=false causes aerosol non-conservation and allows stale work-array values to impact the model state
 #566"
https://github.com/eagles-project/mam4xx/issues/566

It seems one possible fix is to disable one update which would have been a trivial change. But once that line of code was conditional, all the inputs to the line became conditional and it snowballed until one hundred lines were all conditional at which point is seemed better to put all of that code into a function which is then conditionally called.

The fix was not explicitly called out in the issue ticket, but I think this is one possible choice.

### DIFF
--- a/src/mam4xx/ndrop.hpp
+++ b/src/mam4xx/ndrop.hpp
@@ -1159,7 +1159,7 @@ Real explmix(const Real qold_km1, // number / mass mixing ratio from previous
 } // end explmix
 
 KOKKOS_INLINE_FUNCTION
-void aero_vertical_mix(
+void apply_aero_vertical_mixing(
     const ThreadTeam &team, const Real dtmin,
     const Real dtmicro, // time step for microphysics [s]
     const ColumnView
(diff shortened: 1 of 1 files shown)


src/mam4xx/ndrop.hpp
```diff
@@ -1159,8 +1159,8 @@ Real explmix(const Real qold_km1, // number / mass mixing ratio from previous
 } // end explmix
 
 KOKKOS_INLINE_FUNCTION
-void update_from_explmix(
-    const ThreadTeam &team,
+void aero_vertical_mix(
```
jeff-cohere: I like this fix. Can we call this new function `apply_aero_vertical_mixing`?

jeff-cohere (approved):
Worth trying out! It might not be practical, but I wonder if we can add some sort of crude verification test for this, since we have a gross idea of what flipping the switch should do.

odiazib (approved):
It looks good to me. Before merging, I wonder if we should open a PR on the E3SM repo to make sure that this PR, and the previous one that was merged into main, do not produce diffs and that MAM4xx compiles fine with the rest of the E3SM code.

PR #570, opened 2026-07-27
Remove functions: water_uptake::get_e3sm_parameters and ndrop::get_e3sm_parameters.

overfelt:
These functions only copied constant 1D arrays into 2D  arrays.  Just create const static 2D arrays and remove these functions. The new arrays are in aero_config.hpp.

What I thought would be a quick fix became much more extensive.  These static arrays were passed down through multiple layers of functions and on the GPU const static are not allowed to be used this way.  Anyway, there is no need to pass around const static data when the values can just be fetched when needed from anywhere.  

There are lots of files touched but also function signatures simplified and I think the code is a bit cleaner now.

Since just changed how const static data is fetched:
[BFB]

### DIFF
--- a/src/mam4xx/aero_config.hpp
+++ b/src/mam4xx/aero_config.hpp
@@ -73,6 +73,49 @@ public:
 
   /// Aerosol species
   AeroSpeciesView aero_species;
+
+  static constexpr int maxd_aspectype = 14;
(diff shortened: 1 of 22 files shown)


jeff-cohere (approved):

odiazib (approved):
It looks good to me.

PR #576, opened 2026-08-12
Fix entry in index array that was off by one.

overfelt:


odiazib (commented):
I am testing the branch in PR https://github.com/E3SM-Project/E3SM/pull/8620. If all tests pass, we can merge it into main.

odiazib (approved):
Thanks @overfelt!

PR #577, opened 2026-08-18
Expose ndrop mixing substep count

susburrows:
Expose the exact ndrop explicit mixing substep count without changing the physics.

## Summary

- Add an `int& nsubmix` output to `update_from_explmix` and `dropmixnuc`.
- Return the same integer that directly controls the existing explicit mixing loop.
- Update all in-tree validation call sites with local sink variables, leaving their external schemas unchanged.
- Add a deterministic Kokkos device test for the returned count.

## Validation

- `ctest -R '^mam4_ndrop_unit_tests$' --output-on-failure`: 1/1 passed.
- Complete `ndrop`-label suite: 78 passed and 2 existing max-saturation comparisons were skipped by their configured tolerance rule; no numerical failures.
- Independent review: PASS for draft-PR readiness.

## Scope

This change does not alter the computed substep count, loop bounds, equations, limiters, state updates, or production physics.

### DIFF
--- a/src/tests/mam4_ndrop_unit_tests.cpp
+++ b/src/tests/mam4_ndrop_unit_tests.cpp
@@ -1,11 +1,19 @@
 #include <mam4xx/mam4.hpp>
 
+#include "testing.hpp"
+
 #include <catch2/catch.hpp>
 #include <ekat_comm.hpp>
 #include <ekat_logger.hpp>
 #include <ekat_pack_kokkos.hpp>
 
 using mam4::Real;
+
+#ifdef MAM4XX_ENABLE_GPU
+constexpr int team_size = mam4::nlev;
+#else
+constexpr int team_size = 1;
+#endif
 
 // NOTE: other than the nonnegative-ish requirements, this test is basically
 // vacuous, since it mostly does the same thing as the function, but I suppose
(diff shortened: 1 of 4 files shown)


susburrows:
The purpose of this PR is to expose the number of ndrop explicit substeps for diagnostic output, and to implement a namelist switch that bundles this together with other diagnostic output, to assist with understanding the performance impact of the explicit solver.

Developed with codex.  FYI @odiazib @jeff-cohere @overfelt

susburrows:
Used in https://github.com/E3SM-Project/E3SM/pull/8645

src/tests/mam4_ndrop_unit_tests.cpp
```diff
@@ -207,6 +215,61 @@ TEST_CASE("test_explmix", "mam4_ndrop") {
   REQUIRE(mam4::FloatingPoint<Real>::equiv(q, 0.9));
 }
 
+TEST_CASE("test_update_from_explmix_reports_nsubmix", "mam4_ndrop") {
```
odiazib, line 218: I don’t think we need a unit test for these diagnostic variables. While agentic coding makes it easy to add these tests, we also need to consider the long-term maintenance effort.

susburrows: That makes sense.  I can remove it and push a revised version.

jeff-cohere:
@susburrows , [here](https://github.com/eagles-project/mam4xx/blob/main/CMakeLists.txt#L7) are a few simple "on/off" compile-time options we make available for mam4xx. You could add another one for a diagnostics-heavy eval-facing configuration. Basically:

1. Add another `option` alongside the existing ones (e.g. `MAM4XX_EVAL_DIAGNOSTICS`)
2. Edit [mam4_config.hpp.in](https://github.com/eagles-project/mam4xx/blob/main/src/mam4xx/mam4_config.hpp.in) and add a line to pass the CMake option to the preprocessor (e.g. `#cmakedefine MAM4XX_EVAL_DIAGNOSTICS`)
3. Use the macro (`MAM4XX_EVAL_DIAGNOSTICS`) to enable/disable code as needed.

jeff-cohere (approved):
Yep, that looks great!

odiazib:
@jeff-cohere @susburrows, I was looking at the git log for this branch and noticed this commit: [Merge branch 'E3SM-Project:main' into codex/instrument-ndrop-nsubmix](https://github.com/E3SM-Project/mam4xx/pull/577/changes/c5380878907ab07fe5c349649d356a00f8b083b2)

It looks a little strange because of E3SM-Project:main. I usually rebase onto main instead of merging. I wonder whether we should clean this up, i.e., remove this merge commit and rebase instead. On the other hand, I’m not sure whether this is something we need to worry about, and we may be fine just leaving the commit as it is.

jeff-cohere:
I'm not too worried about merge commits, so I think one is fine. @susburrows , are you comfortable with git rebasing? That's how we usually do things to keep the history from getting cluttered with merge commits. Let us know if you'd like to go over the process at some point.

susburrows:
Sure, I can rebase it to clean up the git history.

susburrows:
> Sure, I can rebase it to clean up the git history.

done

jeff-cohere:
Are we ready to merge this one? The one thing I'll say is that, as we add diagnostics to this "eval diagnostics" set, we should probably place them into a struct or something so we're not constantly changing the signatures of functions that deal with them. But we can do this the next time we add such a diagnostic.

susburrows:
I think @odiazib wanted to do some test simulations first

odiazib:
> I think @odiazib wanted to do some test simulations first

@susburrows, here are the results of one test on Chrysalis: /lcrc/group/e3sm/ac.odiazib/test_pr_8645/F2010-EAMxx-MAM4xx_ne30pg2_ne30pg2_L72_mam4xx_pr_8645/run

I can produce a few plots, but I can do that Friday morning. I need to work on another project this Thursday (09/24/2026).

The new debug variables are stored in this file: F2010-EAMxx-MAM4xx_ne30pg2_ne30pg2_L72_mam4xx_pr_8645.dbg.scream.AVERAGE.ndays_x1.0001-01-01-00000.nc

We can also merge this PR, but let’s make sure that the E3SM PR is BFB first.

CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ option(MAM4XX_ENABLE_GPU       "Enable GPU support" OFF)
 option(MAM4XX_ENABLE_COVERAGE  "Enable code coverage instrumentation" OFF)
 option(MAM4XX_ENABLE_SKYWALKER "Enable Skywalker cross validation" ON)
 option(MAM4XX_ENABLE_TESTS     "Enable unit tests" ON)
+option(MAM4XX_EVAL_DIAGNOSTICS "Enable evaluation diagnostics" OFF)
```
odiazib: We should add to the PR that, to enable these variables in E3SM runs, the following command is needed:
`./xmlchange --append SCREAM_CMAKE_OPTIONS="MAM4XX_EVAL_DIAGNOSTICS ON"`

odiazib (approved):

CMakeLists.txt
susburrows: added this to the PR description.

